PR #9, opened 2024-02-19
Multiple Validators Per Safe

seongyun-ko:
# Why
Currently, one safe (EtherFiNode + EigenPod, if restaked) contract need to be minted per validator, which leads to serious gas cost issue.

# What
This PR allows one safe contract to be associated with the multiple validators as long as the validators' (B-NFT holder, T-NFT holder, Node Operator) are the same.

Say a safe contract is associated with one validator `v1`. 

A new validator `v2` can be registered to the same safe if they share the same (B,T NFT holders, Node Operator) entities. The validator `deposit` function is extended to receive an additional parameter `_validatorIdToShareSafeWith`. In this example, when depositing for `v2`, `_validatorIdToShareSafeWith` = `v2`. If `_validatorIdToShareSafeWith` = `0`, it will allocate a new safe.

Now the safe is associated with two validators: `v1` and `v2`. The earned staking rewards are sent to the safe contract and socialized. Either `paritalWithdraw(v1)` or `partialWithdraw(v2)` withdraws the entire staking rewards in one shot. `fullWithdraw(v1)` will withdraw only the principal amount for `v1`.  For the detailed examples, refer to `test_partialWithdrawWithMultipleValidators` in `test/EtherFiNode.t.sol`

Here are some notable points:
- Say there is a slashed validator for a safe, the staking rewards by the other validators will be used to cover the loss of the slashed validator before being paid out as the rewards.
- The full withdrawal only for the last validator will withdraw the staking rewards and principal amount together. 

# Important functions
- `EtherFiNodesManager`'s {`partialWithdraw`,`fullWithdraw`, `processNodeExit`}
- `EtherFiNode`'s {`getRewardsPayouts`, `getFullWithdrawalPayouts`, `getTvlSplits`, `calculateTVL`}


src/StakingManager.sol
```diff
@@ -230,24 +245,34 @@ contract StakingManager is
 
             _cancelDeposit(_validatorIds[x], _caller);
         }
+    }
+
+    /// @dev create a new proxy instance of the etherFiNode withdrawal safe contract.
+    /// @param _createEigenPod whether or not to create an associated eigenPod contract.
+    function instantiateEtherFiNode(bool _createEigenPod) external returns (address) {
```
seongyun-ko, line 252: just moved from `EtherFiNodesManager`

src/EtherFiNodesManager.sol
```diff
@@ -358,6 +359,10 @@ contract EtherFiNodesManager is
             phase: IEtherFiNode(etherfiNode).DEPRECATED_phase()
         });
         IEtherFiNode(etherfiNode).migrateVersion(_validatorId);
+
+        if (_validatorId != 0) {
```
seongyun-ko, line 363: this is the fix for the bug I found.

Without it, the `fullWithdraw` of the validator with its safe version 0 will fail

seongyun-ko:
Audited by Zellic

PR #29, opened 2024-04-01
[DO NOT MERGE] Example: Dummy ETH IOU support for L2 instant minting

JorgeAtPaladin:
@seongyun-ko asked us to demonstrate the changes necessary to introduce instant L2 minting of eETH. The idea would be to have the liquifier accept "dummy ETH" which are essentially predeposit coins that will be swapped against real ETH after the L2 ETH goes through the 7 day bridge.

The advantage of a setup like this is that we can quickly notify the liquifier/deposit pool of new L2 deposits, while bridging the actual ETH over the much slower 7 day bridge.

**DISCLAIMER: THIS IS AN EXAMPLE!** I used the github editor to demonstrate the changes, and did not write or run any tests. When actually introducing this feature, we recommend also to not use the `admin` role for this swap. The Sync contract only needs to call these specific functions, there's no point in giving it excess privileges!

### DIFF
--- a/src/Liquifier.sol
+++ b/src/Liquifier.sol
@@ -41,6 +41,8 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
     IwBETH public wbEth;
     ILido public lido;
 
+    mapping(address => bool) public isDummyToken;
+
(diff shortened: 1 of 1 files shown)


seongyun-ko:
Nice! I made the new one based on it @JorgeAtPaladin 
#30

PR #48, opened 2024-04-23
updated the scripts to ship OFTs to {Base, Linea, BNB}

seongyun-ko:


solipsis (approved):
why all the deletions from forge-std?

seongyun-ko:
> why all the deletions from forge-std?

We had the copy of forge-std library not as a submodule.

PR #50, opened 2024-04-25
feat: adding transaction builder script

jtfirek:
adds the script that generates a json for the transaction builder to resolve:
https://linear.app/ether-fi/issue/ETH-2975/[sc]-generate-the-gnosis-txn-for-timelock-contract

- make sure you have `jq` installed 
- a json file called `transactionBatches.json` is generated in `release/`


seongyun-ko (changes requested):
Thx!
- Let's get rid of these testing log files. you can update `.gitignore` to do this
- Let's add the blocknumber to the output  file name`(blocknumber).transactionBatch.json` ?

seongyun-ko (approved):
Awesome! Appreciate it

PR #56, opened 2024-05-08
Added Permission to Multicall3 contract

vvalecha519:


seongyun-ko (changes requested):
- let's use the latest solidity version
- the code is borrowed from the external. If possible, let's import their one as a library and write ours as a wrapper contract adding the permission control
- did we check the license term?

PR #57, opened 2024-05-09
Fix missed vm.startPrank in tests

PassKeyRa:
Hello team! During tests execution I encountered the following error in `setUpTests` of the `TestSetup.sol` file:

image

Seems that on the [line](https://github.com/etherfi-protocol/smart-contracts/blob/4160ada6203f83e7cf776cc94eb3bb8eeccf0076/test/TestSetup.sol#L663) the `vm.startPrank(owner);` call missed to use the same owner, which deployed the contracts. This PR adds the line


seongyun-ko (approved):
true! thx

PR #58, opened 2024-05-11
withdrawal via the proofs 

seongyun-ko:


src/EtherFiNode.sol
```diff
@@ -34,6 +35,10 @@ contract EtherFiNode is IEtherFiNode {
     mapping(uint256 => uint256) public associatedValidatorIndices;
     uint256[] public associatedValidatorIds;
 
+    // Track the amount of pending/completed withdrawals;
+    uint64 public pendingWithdrawalFromRestakingInGwei; // incremented when the delayed withdrawal (from EigenPod to EtherFiNode) is queued, decremented when it is completed
```
solipsis, line 39: I don't think this will work because anyone can pay the gas to trigger eigenlayer withdrawals. This could cause these variables to be out of sync with reality

PR #61, opened 2024-05-31
add 'seizeInvalidRequest' to WithdrawRequestNFT

seongyun-ko:
Currently, the protocol can invalidate a certain withdraw request NFT and pause its claim.

Say, an NFT was hacked and we did all investigation and KYC.
Then, the protocol has a power to seize the invalidated withdraw request.
We can give the ETH back to the original user

- add 'seizeInvalidRequest'. It withdraws the ETH in the NFT to a recipient.
- the invalidated NFT is now non-transferrable


solipsis (approved):

seongyun-ko:
audit reviewed by Nethermind

PR #66, opened 2024-06-12
Rate Limit on cross-chain transfer from other chains to ETH Mainnet

seongyun-ko:
To limit the risk exposure to malicious outcomes that can happen on bridge,
we put the rate limiter on the cross-chain transfer from other chains to ETH Mainnet.

### DIFF
--- a/src/WeETH.sol
+++ b/src/WeETH.sol
@@ -9,6 +9,9 @@ import "./interfaces/IeETH.sol";
 import "./interfaces/ILiquidityPool.sol";
 import "./interfaces/IRateProvider.sol";
 
+import "src/interfaces/IRateLimiter.sol";
+
+
 contract WeETH is ERC20Upgradeable, UUPSUpgradeable, OwnableUpgradeable, ERC20PermitUpgradeable, IRateProvider {
     //--------------------------------------------------------------------------------------
     //---------------------------------  STATE-VARIABLES  ----------------------------------
@@ -16,6 +19,8 @@ contract WeETH is ERC20Upgradeable, UUPSUpgradeable, OwnableUpgradeable, ERC20Pe
 
     IeETH public eETH;
     ILiquidityPool public liquidityPool;
+    address public oftAdapter;
+    IRateLimiter public rateLimiterOnOftAdapter;
 
     //--------------------------------------------------------------------------------------
     //----------------------------  STATE-CHANGING FUNCTIONS  ------------------------------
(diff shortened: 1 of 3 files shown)


jtfirek:
What are you thoughts on potentially changing the design to instead of having a single rate limiter for all L2s having individual rate limits for each L2? This could be implemented by inheriting the depleting individual rate limiters for each `_origin.srcEid` on the `lzReceive` below:

```
    function lzReceive(
        Origin calldata _origin,
        bytes32 _guid,
        bytes calldata _message,
        address _executor,
        bytes calldata _extraData
    ) public payable virtual {
        // Ensures that only the endpoint can attempt to lzReceive() messages to this OApp.
        if (address(endpoint) != msg.sender) revert OnlyEndpoint(msg.sender);
        
        
        _checkAndUpdateRateLimit(_origin.srcEid, _toLD(_message.amountSD()))
        
        // Ensure that the sender matches the expected peer for the source endpoint.
        if (_getPeerOrRevert(_origin.srcEid) != _origin.sender) revert OnlyPeer(_origin.srcEid, _origin.sender);

        // Call the internal OApp implementation of lzReceive.
        _lzReceive(_origin, _guid, _message, _executor, _extraData);
    }
```

jtfirek:
f well now that I know that `EtherFiOFTAdapter` is not upgradable fgtm

jtfirek (approved):

seongyun-ko:
> What are your thoughts on potentially changing the design to have individual rate limits for each L2, instead of a single rate limiter for all L2s? This could be implemented by inheriting from `@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/utils/RateLimiter.sol` and adding the following `_checkAndUpdateRateLimit` call to the `EtherFiOFTAdapter.sol`.
> 
> ```
>     function lzReceive(
>         Origin calldata _origin,
>         bytes32 _guid,
>         bytes calldata _message,
>         address _executor,
>         bytes calldata _extraData
>     ) public payable virtual {
>         if (address(endpoint) != msg.sender) revert OnlyEndpoint(msg.sender);
>         
>         // checks the rate limit for this inbound chain
>         _checkAndUpdateRateLimit(_origin.srcEid, _toLD(_message.amountSD()))
>         
>         if (_getPeerOrRevert(_origin.srcEid) != _origin.sender) revert OnlyPeer(_origin.srcEid, _origin.sender);
>         _lzReceive(_origin, _guid, _message, _executor, _extraData);
>     }
> ```
> 
> As we add more L2s, the single rate limiter could potentially become overwhelmed with all the inbound messages. We expect to see a large total of all inbound transfers from many different chains, but we need to prevent an single large inbound transfer from a compromised L2.
> 
> Additionally, without matching pairwise rate limit configurations between the source and destination, it is likely that we will see transactions fail in flight after passing the send rate limiter and being blocked at the destination. These transactions will have to be manually unstuck by us on LayerZero scan.
> 
> If we choose to use `@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/utils/RateLimiter.sol` to have pairwise rate limiting for each edge in our mesh network, we will have to edit and audit the contract to store both inbound and outbound rate limits for each `eid`.

Unfortunately, OFT adapter is not upgradeable and we would need to migrate to a new one for this. AFter than, yes 100%

seongyun-ko:
Dropping this since we will migrate OFT adapter to an upgradeable adapter

PR #67, opened 2024-06-19
Eigenlayer test fixes

solipsis:
Most eigenlayer tests working again. For historical proofs need to point at custom rpc that has the historical state at the time of the proofs. More work to be done here but timeboxing the amount spent right now


seongyun-ko (approved):
lgtm

PR #77, opened 2024-06-28
[EFIP-5] Security Improvements for eETH/weETH

seongyun-ko:
## Summary

This EFIP aims to improve the security of ether.fi eETH/weETH by introducing a whitelist on `spender` of `permit` and a blacklist on `transfer`.

## Motivation

While ERC20's extension on `Permit` brings convenience by using off-chain signatures for authorization, it is a well-known vulnerability that can lead to [phishing attacks](https://cointelegraph.com/magazine/phishing-crypto-erc-20-bait-scammers/). We have also been seeing growing incidents of incidents, where eETH/weETH users are tricked into signing a permit for a malicious contract or address.

In addition, the funds can be sent to the malicious contracts or addresses by the owner or the malicious contracts can lock the funds by transferring them to the malicious addresses. This can lead to the loss of funds or the funds being locked forever.

### DIFF
--- a/proposals/EFIP-5.md
+++ b/proposals/EFIP-5.md
@@ -49,7 +49,7 @@ contract EETH {
 
 ### Blacklist on `transfer`
 
-It introduces a blacklist mechanism on `transfer`. It will prevent the transfer of tokens to the blacklisted addresses. This will help to prevent the funds from being sent to the malicious contracts or addresses.
+It introduces a blacklist mechanism on `transfer`. It will prevent the transfer of tokens to and from blacklisted addresses. This will help to prevent the funds from being sent to the malicious contracts or addresses.
 
 Note that [USDCv2](https://etherscan.io/address/0x43506849d7c04f9138d1a2050bbf3a0c054402dd#code) by Circle implements the blacklist.
 
--- a/proposals/EFIP-5.md
+++ b/proposals/EFIP-5.md
@@ -22,7 +22,7 @@ The below are PoC implementations. The actual implementation may vary.
 
 ### Whitelist on `spender` of `permit`
 
-It introduces a whitelist mechanism on `spender` of `permit`. It will only allow the permit to be signed only for the whitelisted addresses to be the `spender` such as the well-known DEXes or lending protocols.
+It introduces a whitelist mechanism on `spender` of `permit`. It will only allow whitelisted addresses to be the `spender` such as the well-known DEXes or lending protocols.
 
 ```
 contract EETH {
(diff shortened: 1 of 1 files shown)


proposals/EFIP-5.md
```diff
@@ -0,0 +1,78 @@
...
+In addition, the funds can be sent to the malicious contracts or addresses by the owner or the malicious contracts can lock the funds by transferring them to the malicious addresses. This can lead to the loss of funds or the funds being locked forever.
+
+
+## Proposal
+
+The below are PoC implementations. The actual implementation may vary.
+
+### Whitelist on `spender` of `permit`
+
+It introduces a whitelist mechanism on `spender` of `permit`. It will only allow the permit to be signed only for the whitelisted addresses to be the `spender` such as the well-known DEXes or lending protocols.
```
solipsis, line 25: -> "It introduces a whitelist mechanism on `spender` of `permit`.  It will only allow whitelisted addresses to be the `spender` such as the well-known DEXes or lending protocols."

Slight change of wording because it doesn't impose restriction on who can sign a permit, only the spender

proposals/EFIP-5.md
```diff
@@ -0,0 +1,78 @@
...
+
+        ...
+    }
+}
+```
+
+
+### Blacklist on `transfer`
+
+It introduces a blacklist mechanism on `transfer`. It will prevent the transfer of tokens to the blacklisted addresses. This will help to prevent the funds from being sent to the malicious contracts or addresses.
```
solipsis, line 52: "It will prevent the transfer of tokens to and from blacklisted addresses"

Mechanism is on both sending and receiving side

proposals/EFIP-5.md
```diff
@@ -0,0 +1,78 @@
...
+Note that [USDCv2](https://etherscan.io/address/0x43506849d7c04f9138d1a2050bbf3a0c054402dd#code) by Circle implements the blacklist.
+
+```
+contract EETH {
+    ...
+    
+    mapping(address => bool) public blacklisted;
+
+    function _transfer(address _sender, address _recipient, uint256 _amount) internal {
+        require(!blacklisted[_sender] || !blacklisted[_recipient], "EETH: blacklisted address");
```
solipsis, line 63: I don't think this condition is quite what you want. I think you want `&&`. As written you are fine as long as both sender and recipient are not both blacklisted

solipsis (changes requested):
Couple small wording recommendations, and I believe an issue on your blacklist condition

proposals/EFIP-5.md
seongyun-ko: yes, && thx!

solipsis (approved):

PR #81, opened 2024-07-03
EtherFiStaking with delegation/voting

seongyun-ko:
A simple ETHFI staking contract :)

- support delegation of voting via {`deposit`, `delegate`}. the delegation can happen during the staking together.
- support voting via Snapshot.org


solipsis:
since this is a new project, do we want to put it in a separate repo like we have for the other recent contracts so that we can utilize OZ version 5.x and other new features?

seongyun-ko:
> since this is a new project, do we want to put it in a separate repo like we have for the other recent contracts so that we can utilize OZ version 5.x and other new features?

Is it not possible to upgrade from OZ ver <5 to ver 5.x ?

seongyun-ko:
we are not going with it. bye bye

PR #82, opened 2024-07-05
[Implementation][EFIP-11] Async Admin Task Execution for Validator Management

seongyun-ko:
- `executeTasks` now enqueues the tasks for admin tasks so that they can be processed async instead of sync processing them in a single txn
- `_enqueueAdminTask_ApproveValidators` to split the tasks and queue them
- `AdminTask_ApproveValidators` to process the pending admin tasks in LIFO manner

Any feedback, extension welcome


src/EtherFiAdmin.sol
```diff
@@ -164,6 +170,31 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+
+        require(etherFiOracle.isConsensusReached(_reportHash), "EtherFiAdmin: report didn't reach consensus");
+        require(task.operationHash == keccak256(abi.encode(_reportHash, _validatorsToApprove, _pubKey, _signature)), "EtherFiAdmin: invalid task");
+
+        liquidityPool.batchApproveRegistration(_validatorsToApprove, _pubKey, _signature);
+    }
+
+    // FIFO is better?
+    // should we support random-access so that the admin can process the pending transactions in any order?
+    function _dequeuePendingTask(string memory _operation) internal returns (AdminTask memory) {
```
vvalecha519, line 185: maybe this should be an external function with onlyAdmin modifier. What happens if admin wants to unpublished? How do we clear the queue?

seongyun-ko: good point. we need to add a function to handle such cases

src/EtherFiAdmin.sol
```diff
@@ -153,7 +159,7 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         numValidatorsToSpinUp = _report.numValidatorsToSpinUp;
 
         _handleAccruedRewards(_report);
-        _handleValidators(_report, _pubKey, _signature);
+        _handleValidators(reportHash, _report, _pubKey, _signature);
```
vvalecha519, line 162: should we remove _pubKey in _handleValidators waste of massive argument. It does nothing but get hashed in the _enqueueAdminTask_ApproveValidators. The benefit is that during the same pubkeys are provided as were passed during the executeTask, but if thats wrong then this is wrong as well.

vvalecha519 (commented):
Should executeTask revert if the approving validator queue is empty? Otherwise lastHandledReportRefSlot is wrong because the validators are not all proccessed but we claim the reported has been handled

src/EtherFiAdmin.sol
```diff
@@ -189,25 +220,48 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
         if (_report.validatorsToApprove.length > 0) {
-            liquidityPool.batchApproveRegistration(_report.validatorsToApprove, _pubKey, _signature);
+            _enqueueAdminTask_ApproveValidators(_reportHash, _report.validatorsToApprove,  _pubKey, _signature);
+            // liquidityPool.batchApproveRegistration(_report.validatorsToApprove, _pubKey, _signature);
         }
 
         // liquidityPoolValidatorsToExit
         if (_report.liquidityPoolValidatorsToExit.length > 0) {
-            liquidityPool.sendExitRequests(_report.liquidityPoolValidatorsToExit);
+            // liquidityPool.sendExitRequests(_report.liquidityPoolValidatorsToExit);
```
vvalecha519, line 254: why did we get rid of this?

vvalecha519:
updated setBatchSize to setValidatorTaskBatchSize as you recommended

vvalecha519 (approved):

PR #83, opened 2024-07-05
[Implementation][EFIP-4] processing treasury and NO fee in eeth

vvalecha519:


src/LiquidityPool.sol
```diff
@@ -42,6 +43,7 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, IL
     uint128 public totalValueInLp;
 
     address public DEPRECATED_admin;
+    address public treasury;
```
seongyun-ko, line 46: Adding new state variables in the upgradeable contract must be done in append-only manner only.
https://docs.openzeppelin.com/learn/upgrading-smart-contracts#upgrading

But in this case, I think we can re-purpose the deprecated variable `DEPRECATED_admin` for `treasury` :)

src/LiquidityPool.sol
```diff
@@ -462,10 +464,27 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, IL
         emit Rebase(getTotalPooledEther(), eETH.totalShares());
     }
 
+    function payProtocolFees(int128 _protocolFees) public {
```
seongyun-ko, line 467: Unless we consider particular cases of penalizing the protocol (we dont have it as of now at least)
we don't need to consider the negative protocol fee which leads to `burnShares` in the below flow.

For now, let's make it simple and consider only positive protocol fee flows.

If any slashing happens and we have to slash eETH, `_protocolFees` will be `0`.
Slashing on eETH will penalize the protocol's eETH holdings equally, anyway

test/EtherFiOracle.t.sol
seongyun-ko, line 1: @vvalecha519 we need test cases for `payProtocolFees` when `_protocolFees > 0`.
+ write some detailed scenarios around it.

Assum 100 ETH is earned from ether.fi validators, how should it be split to {protocolAccruedRewards, protocolFees}

vvalecha519: will do

seongyun-ko:
can we just let `Rebase` mint the fee to the treasury directly?

src/LiquidityPool.sol
```diff
@@ -462,10 +464,21 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, IL
         emit Rebase(getTotalPooledEther(), eETH.totalShares());
     }
 
+    function payProtocolFees(uint128 _protocolFees) public {
+        if (msg.sender != address(etherFiAdminContract)) revert IncorrectCaller();
+        uint256 treasuryShares = sharesForAmount(_protocolFees);
+        totalValueOutOfLp = uint128(int128(totalValueOutOfLp) + _protocolFees);
+        eETH.mintShares(treasury, treasuryShares);
+    }
```
seongyun-ko, line 472: I think `totalValueOutOfLp` must be calculated first, and then `treasuryShares`

plz verify with tests

```
    function payProtocolFees(uint128 _protocolFees) public {
        if (msg.sender != address(etherFiAdminContract)) revert IncorrectCaller();
        totalValueOutOfLp = uint128(int128(totalValueOutOfLp) + _protocolFees);
        uint256 treasuryShares = sharesForAmount(_protocolFees);
        eETH.mintShares(treasury, treasuryShares);
    }
```

test/EtherFiOracle.t.sol
```diff
@@ -648,13 +648,33 @@ contract EtherFiOracleTest is TestSetup {
...
+        IEtherFiOracle.OracleReport memory report = _emptyOracleReport();
+        _initReportBlockStamp(report);
+
+        // Assume that the accruedRewards must be 1 ether, all the time
+
+        // Alice submited the correct report 
+        vm.prank(alice);
+        report.protocolAccruedRewards = 90 ether;
+        report.protocolFees = 10 ether;
+    }
```
seongyun-ko, line 679: I believe the test is not complete? WIP?

src/LiquidityPool.sol
vvalecha519: oohhhh good catch

vvalecha519:
> Rebase

Rebase and minting shares for protocol fees seem unrelated I believe we should just add a new method

src/LiquidityPool.sol
solipsis: I see this conversation is marked resolved, but Sykos point has not been addressed in this commit

src/LiquidityPool.sol
```diff
@@ -462,10 +464,38 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, IL
         emit Rebase(getTotalPooledEther(), eETH.totalShares());
     }
 
+    function payProtocolFees(uint128 _protocolFees) public {
+        if (msg.sender != address(etherFiAdminContract) && msg.sender != address(this)) revert IncorrectCaller();
```
solipsis, line 468: this `msg.sender != address(this)` is a littly janky

I would make `_payProtocolFees` an internal function
And have an external version 
```
payProtocolFees() external { 
  _payProtocolFees()
}
```
And call the internal version from the other function that calls this method. And this would match how he handle this situation in our other contracts

src/LiquidityPool.sol
```diff
@@ -462,10 +464,38 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, IL
...
+        eETH.mintShares(treasury, treasuryShares);
+    }
+
+    //one time function to mint shares when we change split to 0
+    //need to 
+    function mintShareOnChangeSplit(uint256[] memory _validatorIds, uint256[] memory _beaconBalances) external onlyAdmin { 
+            uint256 totalAccruedRewards = 0;
+            for (uint256 i = 0; i < _validatorIds.length; i++) {
+                require(_beaconBalances[i] <= 33 ether, "Invalid Beacon Balance");
+                (uint256 treasury, uint256 nodeOperator, uint256 tnft, uint256 bnft) = nodesManager.calculateTVL(_validatorIds[i], _beaconBalances[i]);
```
solipsis, line 480: I believe calculateTVL is currently broken in the current implementation, so we need to be very careful using it here to handle payouts (It does not properly account for funds in eigenlayer `delegationManager` withdrawal queue

test/TestSetup.sol
```diff
@@ -714,15 +714,15 @@ contract TestSetup is Test {
...
-        reportAtPeriod3B = IEtherFiOracle.OracleReport(1, 0, 2048 - 1, 1, 2 * 1024 - 1, 1, validatorsToApprove, validatorsToExit, exitedValidators, exitTimestamps, slashedValidators, withdrawalRequestsToInvalidate, 1, 80, 20, 0, 0);
-        reportAtPeriod4 = IEtherFiOracle.OracleReport(1, 2 * 1024, 1024 * 3 - 1, 2 * 1024, 3 * 1024 - 1, 0, validatorsToApprove, validatorsToExit, exitedValidators, exitTimestamps, slashedValidators, withdrawalRequestsToInvalidate, 1, 80, 20, 0, 0);
-        reportAtSlot3071 = IEtherFiOracle.OracleReport(1, 2048, 3072 - 1, 2048, 3072 - 1, 1, validatorsToApprove, validatorsToExit, exitedValidators, exitTimestamps, slashedValidators, withdrawalRequestsToInvalidate, 1, 80, 20, 0, 0);
-        reportAtSlot4287 = IEtherFiOracle.OracleReport(1, 3264, 4288 - 1, 3264, 4288 - 1, 1, validatorsToApprove, validatorsToExit, exitedValidators, exitTimestamps, slashedValidators, withdrawalRequestsToInvalidate, 1, 80, 20, 0, 0);
+        reportAtPeriod2A = IEtherFiOracle.OracleReport(1, 0, 1024 - 1, 0, 1024 - 1, 1, 0,validatorsToApprove, validatorsToExit, exitedValidators, exitTimestamps, slashedValidators, withdrawalRequestsToInvalidate, 1, 80, 20, 0, 0);
```
solipsis, line 717: would it be possible to update this test to have some more named variables? At a glance I have no idea what each of these numbers represents.

src/LiquidityPool.sol
vvalecha519: how do we update it so it is correct?

src/LiquidityPool.sol
seongyun-ko: @vvalecha519

src/LiquidityPool.sol
seongyun-ko: image

src/LiquidityPool.sol
seongyun-ko: agreed. @vvalecha519 https://docs.soliditylang.org/en/latest/control-structures.html#internal-function-calls

test/EtherFiOracle.t.sol
```diff
@@ -648,13 +648,33 @@ contract EtherFiOracleTest is TestSetup {
...
+
+        IEtherFiOracle.OracleReport memory report = _emptyOracleReport();
+        _initReportBlockStamp(report);
+
+        // Assume that the accruedRewards must be 1 ether, all the time
+
+        // Alice submited the correct report 
+        vm.prank(alice);
+        report.protocolAccruedRewards = 90 ether;
+        report.protocolFees = 10 ether;
```
seongyun-ko, line 678: @vvalecha519 wen test

test/EtherFiOracle.t.sol
seongyun-ko: wen

test/EtherFiOracle.t.sol
seongyun-ko: (Discussion with V)

protocolAccruedRewards = rebase eeth (so includes tnfts eth and bnft eth where bnft owned by liquidity pool)
protocolFees = ethfund eth rewards + treasury rewards + no operator rewards

if 100 eth is earned by etherfi validators 
lets z= fraction of bnft held by lp
protocolAccruedRewards = 90 * 29/32 + 3/32*90*(z)
protocolFees = 10 + 3/32*90 * (1-z)

protocolAccruedRewards + protocolFees = 100

src/interfaces/IEtherFiOracle.sol
```diff
@@ -8,7 +8,8 @@ interface IEtherFiOracle {
         uint32 refSlotTo;
         uint32 refBlockFrom;
         uint32 refBlockTo;
-        int128 accruedRewards;
+        int128 protocolAccruedRewards;
```
seongyun-ko, line 11: let's keep the naming since it is the same thing as `accruedRewards`

src/LiquidityPool.sol
```diff
@@ -9,6 +9,7 @@ import "@openzeppelin-upgradeable/contracts/token/ERC721/IERC721ReceiverUpgradea
 import "@openzeppelin-upgradeable/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgradeable/contracts/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import "forge-std/console.sol";
```
solipsis, line 12: lets drop this when not testing

solipsis (commented):
seems fine. But I will let Syko give the final signoff for this functionality due to my unfamiliarity

test/TestSetup.sol
vvalecha519: Jacob and I can fix this during the refactor

src/EtherFiAdmin.sol
```diff
@@ -164,6 +163,15 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         emit AdminOperationsExecuted(msg.sender, reportHash);
     }
 
+    //protocol owns the eth that was distributed to NO and treasury in eigenpods and etherfinodes 
+    function _handleProtocolFees(IEtherFiOracle.OracleReport calldata _report) internal { 
+        require(_report.protocolFees >= 0, "EtherFiAdmin: protocol fees can't be negative");
```
seongyun-ko, line 168: let's use unsigned int tyep for `protocolFees` then

src/EtherFiAdmin.sol
```diff
@@ -262,7 +270,7 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable {
 
 
     modifier isAdmin() {
-        require(admins[msg.sender] || msg.sender == owner(), "EtherFiAdmin: not an admio");
+        require(admins[msg.sender] || msg.sender == owner(), "EtherFiAdmin: not an admin");
```
seongyun-ko, line 273: @jtfirek will need to visit this later then

test/EtherFiOracle.t.sol
seongyun-ko: @vvalecha519 where is the test for this scenario?

seongyun-ko: you can spin up 2 vals so that z = 1/2 and do the test.

PR #85, opened 2024-07-10
[EFIP-5] {Permit Whitelist, Transfer Blacklist}

jtfirek:
- Implements [EFIP-5](https://github.com/etherfi-protocol/smart-contracts/blob/master/proposals/EFIP-5.md) functionality and tests
- Makes comment style consistent between `eETH` and `weETH`

### DIFF
--- a/src/EETH.sol
+++ b/src/EETH.sol
@@ -158,7 +158,7 @@ contract EETH is IERC20Upgradeable, UUPSUpgradeable, OwnableUpgradeable, IERC20P
     //--------------------------------------------------------------------------------------
 
     function _transfer(address _sender, address _recipient, uint256 _amount) internal {
-        require(blacklistedRecipient[_recipient] == false, "eETH: recipient blacklisted"); 
+        require(!blacklistedRecipient[_sender] && !blacklistedRecipient[_recipient], "eETH: blacklisted address"); 
         
         uint256 _sharesToTransfer = liquidityPool.sharesForAmount(_amount);
         _transferShares(_sender, _recipient, _sharesToTransfer);
--- a/src/EETH.sol
+++ b/src/EETH.sol
@@ -15,12 +15,18 @@ import "./interfaces/ILiquidityPool.sol";
 
 contract EETH is IERC20Upgradeable, UUPSUpgradeable, OwnableUpgradeable, IERC20PermitUpgradeable, IeETH {
     using CountersUpgradeable for CountersUpgradeable.Counter;
+    //--------------------------------------------------------------------------------------
+    //---------------------------------  STATE-VARIABLES  ----------------------------------
+    //--------------------------------------------------------------------------------------
+    
     ILiquidityPool public liquidityPool;
 
     uint256 public totalShares;
     mapping (address => uint256) public shares;
     mapping (address => mapping (address => uint256)) public allowances;
     mapping (address => CountersUpgradeable.Counter) private _nonces;
+    mapping (address => bool) public whitelistedSpender;
+    mapping (address => bool) public blacklistedRecipient;
 
     bytes32 private constant _PERMIT_TYPEHASH = keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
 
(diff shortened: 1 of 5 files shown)


src/EETH.sol
```diff
@@ -15,12 +15,18 @@ import "./interfaces/ILiquidityPool.sol";
...
+    //---------------------------------  STATE-VARIABLES  ----------------------------------
+    //--------------------------------------------------------------------------------------
+    
     ILiquidityPool public liquidityPool;
 
     uint256 public totalShares;
     mapping (address => uint256) public shares;
     mapping (address => mapping (address => uint256)) public allowances;
     mapping (address => CountersUpgradeable.Counter) private _nonces;
+    mapping (address => bool) public whitelistedSpender;
```
jtfirek, line 28: @seongyun-ko, Could you confirm my reasoning here:

Becausing constant and immutable variables are stored directly in the bytecode, I can place the new mapping above these values without having to worry about a storage collision

seongyun-ko: ho, sounds right :) you can always verify it by checking the storage layout of this contract before/after your update

src/EETH.sol
```diff
@@ -139,8 +153,13 @@ contract EETH is IERC20Upgradeable, UUPSUpgradeable, OwnableUpgradeable, IERC20P
         _approve(owner, spender, value);
     }
 
-    // [INTERNAL FUNCTIONS] 
+    //--------------------------------------------------------------------------------------
+    //-------------------------------  INTERNAL FUNCTIONS  ---------------------------------
+    //--------------------------------------------------------------------------------------
+
     function _transfer(address _sender, address _recipient, uint256 _amount) internal {
+        require(blacklistedRecipient[_recipient] == false, "eETH: recipient blacklisted");
```
seongyun-ko, line 161: According to the spec, the check happens  both on sender and recipient; 

```
require(!blacklisted[_sender] || !blacklisted[_recipient], "EETH: blacklisted address");
```

seongyun-ko: Q2) how much gas cost does it add?

src/WeETH.sol
```diff
@@ -76,14 +82,64 @@ contract WeETH is ERC20Upgradeable, UUPSUpgradeable, OwnableUpgradeable, ERC20Pe
...
     ) internal override onlyOwner {}
 
+
+    /// @notice Require the recipient to not be blacklisted before calling {ERC20Upgradeable-_transfer}
+    function _transfer(
+        address from,
+        address to,
+        uint256 amount
+    ) internal override {
+        require(blacklistedRecipient[to] == false, "weETH: recipient blacklisted");
```
seongyun-ko, line 117: the same here

test/TestSetup.sol
```diff
@@ -703,6 +703,12 @@ contract TestSetup is Test {
         _initializeEtherFiAdmin();
 
         admin = alice;
+
+        // weETH Instance must be on eETH whitelist for `weETH.wrapWithPermit` to function as expected
```
seongyun-ko, line 707: kk

seongyun-ko (changes requested):
THX!

src/EETH.sol
jtfirek: Initial Gas Cost: 49948
Gas Cost with check only on `to`: 52161
Gas Cost if we add checks on both `to` and `from`: 54382

src/WeETH.sol
jtfirek: Initial Gas Cost: 29848
Gas Cost with check only on `to`: 32100
Gas Cost if we add checks on both `to` and `from`: 34321

jtfirek:
@solipsis @seongyun-ko up to you guys if you think the extra 2000 gas on the `SLOAD` is worth it to apply the sender as well

src/EETH.sol
seongyun-ko: thx that is good to know.

2100 is due to the cold SLOAD :)
https://hackmd.io/@fvictorio/gas-costs-after-berlin#SLOAD-after-Berlin

src/WeETH.sol
seongyun-ko: thx

seongyun-ko:
> @solipsis @seongyun-ko up to you guys if you think the extra 2000 gas on the `SLOAD` is worth it to apply the transfer blacklist to the sender as wel

yup we should

src/EETH.sol
jtfirek: Verified ✅

seongyun-ko (approved):
awesome LGTM

PR #86, opened 2024-07-11
Fine-Grained Timelock EFIP Proposal

jtfirek:


seongyun-ko:
@solipsis is it what we can add to v2.5?

jtfirek:
> @solipsis is it what we can add to v2.5?

This is fresh contract not an upgrade to an existing so I believe we can use it in this case!

solipsis:
Yes we can use it here. It depends on the relationship graph. We are not able to have our 4.x contracts inherit from `AccessManaged` and use the `AccessManager` as full fledged permission manager. We are able to deploy this and use it as a finegrained timelock to call our existing contracts.

solipsis (approved):

PR #89, opened 2024-07-12
[EFIP-8] Rescue weEth from Treasury

vvalecha519:


seongyun-ko:
Possibly, it can break the downstream dapps
e.g., Say a dapp is performing the accountings on weETH balances within some contracts, for some reasons.
An attacker can break their accounting logics for fun and profits

seongyun-ko:
An example for more secure version only for ether.fi treasury contract

```
function rescueWeEthWithEth(weEthAmount) payable onlyOwner {
  target = etherfi treasury contract
  // is target whitelisted
  // weEth.balanceOf(target) > weEthAmount
  // msg.value >= weEthAmount * exchangeRate()
  _transfer(target, msg.sender);
  target.call{value: msg.value}();
}
```

PR #93, opened 2024-07-17
Deprecate the delegated staking

seongyun-ko:
Back then, anyone could become the owner of a validator by putting their 32 ether.
However, it didn't get much attention from our users.

During the previous upgrade, the feature was already deprecated by adding the flag `isFullStakeEnabled` and setting it to `false`. But the code has remained. This PR removes the code for the deprecated functionality + do some refactoring on the unit tests.

TODO
- Improve the interaction between StakingManager, LiquidityPool when spinning up the validators
- handle TODO in the unit test `test_RegisterOne` -> DONE


src/StakingManager.sol
```diff
@@ -395,6 +340,10 @@ contract StakingManager is
         bytes32 depositDataRoot = depositRootGenerator.generateDepositRoot(_depositData.publicKey, _depositData.signature, withdrawalCredentials, _depositAmount);
         require(depositDataRoot == _depositData.depositDataRoot, "WRONG_ROOT");
 
+        bytes32 fullHash = keccak256(abi.encode(_validatorId, _staker, _tNftRecipient, _bNftRecipient));
+        bytes10 truncatedHash = bytes10(fullHash);
```
solipsis, line 344: by birthday paradox approximation, collisions become likely once we hit 1 trillion validators :)

solipsis (approved):

src/StakingManager.sol
seongyun-ko: hope that happens!

PR #84, opened 2024-07-10
role-registry + granulare EFNM call whitelist

solipsis:
This PR adds a RoleRegistry contract that can serve as the central point of permissions for future updates. It is bundled with an update to switch the EtherfiNodesManager blacklist to a more secure and granular whitelist implementation. I bundled these changes to ensure the role registry was capable of fulfilling its first intended purpose

This PR can serve as a launching point for additional permissions/role based changes for the 2.5 project

* RoleRegistry contract
* call blacklist -> whitelist
* simplified external calling architecture + security
* updated all tests to new system

PR #94, opened 2024-07-18
Jtdev/feat/efip 6

jtfirek:
`Pauser.sol` is a single contract for managing all pausing and unpausing functionality for the protocol. It has a configurable time delay for the execution of unpause actions.

This contract was used as a starting point:
https://github.com/Se7en-Seas/boring-governance/blob/main/src/base/Roles/Pauser.sol

The implementation deviated from the approved proposal document. Should I update the proposal document to match the implementation?

### DIFF
--- a/proposals/EFIP-6.md
+++ b/proposals/EFIP-6.md
@@ -1,0 +1,33 @@
+# [EFIP-6] Fine-Grained Timelock
+
+
+**Author**: jtdev (jacob@ether.fi) 
+
+**Date**: 2000-07-11
+
+## Summary
+
+This proposal suggests implementing a fine-grained timelock with OpenZeppelin Access Control v5 system for HyperNative’s automated response to alerts. Specifically, it proposes reducing the timelock for unpausing smart contracts from 3 days to 15 minutes. This change aims to mitigate the impact of false-positive alerts and allow for more aggressive alert configurations to better protect the protocol.
+
+## Motivation
+
+Currently, when HyperNative pauses our smart contracts (SCs) in response to alerts, it takes 3 days to unpause them. This long delay can be problematic, especially in the case of false-positive alerts. By reducing the unpause timelock to 15 minutes, we can decrease the cost of false-positive alerts and configure the alerts more aggressively
+
+## Proposal
+
+To achieve this, we will deploy the OpenZeppelin Access Control v5 contract to manage all necessary functionality. This contract is very powerful with numerous features including fine-grained timelock functionality.
+
+The existing timelock will remain in place to handle the most critical actions, such as upgrading contracts. However, it can delegate other actions with varying degrees of locking. Specifically, we propose:
+
+	1. Deploy OpenZeppelin Access Control v5
+    2. Configure Access Control contract to have permission for unpausing smart contracts with a 15 minute timelock
+	3. Maintaining the 3-day timelock for other critical actions, such as contract upgrades.
+
+## References
+
+- [Access Control - OpenZeppelin Docs](https://docs.openzeppelin.com/contracts/5.x/api/access#AccessManager)
+- [Access Control v5 - Code](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.1/contracts/access/manager/AccessManager.sol)
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/src/Pauser.sol
+++ b/src/Pauser.sol
@@ -19,26 +19,6 @@ contract Pauser is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     IPausable[] public pausables;
 
     /**
-     * @notice Used to pause calls to `deposit` and `depositWithPermit`.
-     */
-    bool public isPaused;
-
-    /**
-     * @notice Minimum delay required to execute an unpause action.
-     */
-    uint256 public minUnpauseDelay;
-
-    /**
-     * @notice Mapping of unpause operations and the time they become executable.
-     */
-    mapping(bytes32 id => uint256) public unpauseExecutionTime;
-
-    /**
-     * @notice Timestamp indicating an operation has been executed.
-     */
-    uint256 public constant _DONE_TIMESTAMP = uint256(1);
-
-    /**
      * @notice Global Role registry contract.
      */
     RoleRegistry public roleRegistry;
@@ -49,12 +29,6 @@ contract Pauser is Initializable, UUPSUpgradeable, OwnableUpgradeable {
      * @notice Contract specific roles.
      */
     bytes32 public constant PAUSER_ADMIN = keccak256("PAUSER_ADMIN");
-    /**
-     * @notice Protocol specific roles.
-     */
-    bytes32 public constant PROTOCOL_PAUSER = keccak256("PROTOCOL_PAUSER");
-    bytes32 public constant PROTOCOL_UNPAUSER = keccak256("PROTOCOL_UNPAUSER");
-    bytes32 public constant PROTOCOL_UPGRADER = keccak256("PROTOCOL_UPGRADER");
 
     //============================== ERRORS ===============================
 
@@ -67,12 +41,6 @@ contract Pauser is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     event PausableAdded(address indexed pausable);
     event PausableRemoved(address indexed pausable);
 
-    event MinUnpauseDelayUpdated(uint256 oldDuration, uint256 newDuration);
-    event UnpauseScheduled(bytes32 indexed id, uint256 indexed timestampProposed, uint256 timestampExecutable);
-    event UnpauseExecuted(bytes32 indexed id);
-    event UnpauseDeleted(bytes32 indexed id);
-    
-
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
         _disableInitializers();
@@ -84,9 +52,6 @@ contract Pauser is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         for (uint256 i = 0; i < _pausables.length; ++i) {
             pausables.push(_pausables[i]);
         }
-
-        minUnpauseDelay = _minUnpauseDelay;
-        emit MinUnpauseDelayUpdated(0, _minUnpauseDelay);
 
         roleRegistry = RoleRegistry(_roleRegistry);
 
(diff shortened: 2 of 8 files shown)


src/Pauser.sol
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * @notice Global Role registry contract.
+     */
+    RoleRegistry public roleRegistry;
+
+    //============================== ROLES ================================
+
+    bytes32 public constant PAUSER_ADMIN = keccak256("PAUSER_ADMIN");
+    bytes32 public constant PROTOCOL_PAUSER = keccak256("PROTOCOL_PAUSER");
```
jtfirek, line 49: @solipsis Ok that I refined the roles from `RoleRegistry` for easy access?

src/Pauser.sol
```diff
@@ -0,0 +1,320 @@
...
+    //============================== ROLES ================================
+
+    /**
+     * @notice Contract specific roles.
+     */
+    bytes32 public constant PAUSER_ADMIN = keccak256("PAUSER_ADMIN");
+    /**
+     * @notice Protocol specific roles.
+     */
+    bytes32 public constant PROTOCOL_PAUSER = keccak256("PROTOCOL_PAUSER");
```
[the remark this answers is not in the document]
solipsis: not a problem in itself. Would become a tradeoff if this contract started nearing the size limit. Other potential footgun would be a situation where we changed the role on one contract but not the other. Unlikely but something we need to be aware of if you have it in 2 separate places and something we could possibly screw up in a future refactor. Not saying what you did is wrong, just that we are still feeling out this setup since we haven't used it before and we should think through situations like this

test/Pauser.t.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    function setUp() public {
+        setUpTests();
+        
+        // deploy+initialize the pauser
+        Pauser pauserImplementation = new Pauser();
+        initialPausables.push(IPausable(address(liquifierInstance)));
+        initialPausables.push(IPausable(address(etherFiOracleInstance)));
+        
+        bytes memory initializerData = abi.encodeWithSelector(Pauser.initialize.selector, initialPausables, INITIAL_DELAY,  address(roleRegistry));
```
solipsis, line 21: Ty for using the new initialization pattern

src/Pauser.sol
```diff
@@ -0,0 +1,320 @@
...
+    /**
+     * @notice Removes a contract from the list of pausables.
+     */
+    function removePausable(uint256 _index) external onlyRole(PAUSER_ADMIN) {
+        uint256 pausablesLength = pausables.length;
+        if (_index >= pausablesLength) {
+            revert Pauser__IndexOutOfBounds();
+        }
+        address removed = address(pausables[_index]);
+        pausables[_index] = pausables[pausablesLength - 1];
```
solipsis, line 114: while a swap remove is efficient, I think it breaks the semantics you have established elsewhere in the contract. All your events are based around an indexed `index` field. After performing this operation, anyone relying on those events would no longer have a valid view of the world

solipsis: Lets add a convenience function that helps you find this value. 
`indexOf(address)` or something to that effect. Not the most effecient onchain but would be handy offchain and via etherscan when you want to perform a remove operation

solipsis (commented):
Good work. Other question I have for the execs / @seongyun-ko  is if we want any roles that are able to operate outside of the minimum delay you have configured here. Like should super admin gnosis be able to unpause without delay? Just thinking of the 2 or 3 examples we have where we briefly paused something after a hypernative alert and then unpaused a few minutes later

src/Pauser.sol
jtfirek: added ✅

seongyun-ko:
> Good work. Other question I have for the execs / @seongyun-ko is if we want any roles that are able to operate outside of the minimum delay you have configured here. Like should super admin gnosis be able to unpause without delay? Just thinking of the 2 or 3 examples we have where we briefly paused something after a hypernative alert and then unpaused a few minutes later

I think the un-pause can happen without delay

proposals/EFIP-6.md
```diff
@@ -0,0 +1,33 @@
+# [EFIP-6] Fine-Grained Timelock
```
seongyun-ko, line 1: Since the implementation adds a new `pauser` contract as a single entry point for the protocol pause/unpause, you can update the EFIP slightly :)

src/Pauser.sol
```diff
@@ -0,0 +1,332 @@
...
+
+contract Pauser is Initializable, UUPSUpgradeable, OwnableUpgradeable {
+    // ========================================= STATE =========================================
+
+    /**
+     * @notice List of contracts that can be paused and unpaused using:
+     * - `pauseAll`
+     * - `unpauseAll`
+     */
+    IPausable[] public pausables;
```
seongyun-ko, line 19: Do all of our contracts implement`IPausable` ?

src/Pauser.sol
```diff
@@ -0,0 +1,332 @@
...
+            pausables.push(_pausables[i]);
+        }
+
+        minUnpauseDelay = _minUnpauseDelay;
+        emit MinUnpauseDelayUpdated(0, _minUnpauseDelay);
+
+        roleRegistry = RoleRegistry(_roleRegistry);
+
+        __Ownable_init();
+        __UUPSUpgradeable_init();
```
seongyun-ko, line 94: they are better to be initialized first case they are for inherit

src/Pauser.sol
```diff
@@ -0,0 +1,332 @@
...
+     */
+    function updateDelay(uint256 _minDelay) external onlyRole(PAUSER_ADMIN) {
+        minUnpauseDelay = _minDelay;
+        emit MinUnpauseDelayUpdated(minUnpauseDelay, _minDelay);
+    }
+
+    /**
+     * @notice Deletes a scheduled unpause action.
+    */
+    function deleteUnpause(bytes32 _id) external onlyRole(PAUSER_ADMIN) {
```
seongyun-ko, line 130: `cancelUnpause` ?

src/Pauser.sol
```diff
@@ -0,0 +1,332 @@
...
+        uint256 timestamp = unpauseExecutionTime[_id];
+
+        return (timestamp != 0 && timestamp != _DONE_TIMESTAMP && timestamp <= block.timestamp);
+    }
+    
+    /**
+     * @notice Returns the hash id of a single contract unpause with the timestamp it was proposed.
+     */
+    function hashUnpauseSingle(IPausable _pausable, uint256 _timestampProposedProposed) public pure returns (bytes32) {
+        return keccak256(abi.encode(_pausable, _timestampProposedProposed));
```
seongyun-ko, line 298: `_timestampProposed`

test/Pauser.t.sol
```diff
@@ -0,0 +1,147 @@
...
+        roleRegistry.grantRole(pauser.PROTOCOL_UNPAUSER(), alice);
+        roleRegistry.grantRole(pauser.PAUSER_ADMIN(), alice);
+        vm.startPrank(alice);
+    }
+
+    function test_singlePause() public {
+        // pausing the `liquifier` and `etherFiOracle`
+        pauser.pauseSingle(IPausable(address(liquifierInstance)));
+        pauser.pauseSingle(IPausable(address(etherFiOracleInstance)));
+        assertTrue(liquifierInstance.paused());
```
seongyun-ko, line 42: check if `etherFiOracleInstance` is paused too

test/Pauser.t.sol
```diff
@@ -0,0 +1,147 @@
...
+
+        // scheduling the unpause of `liquifier` and warping past the delay
+        pauser.scheduleUnpauseSingle(IPausable(address(liquifierInstance)));
+        assertTrue(liquifierInstance.paused());
+        vm.warp(block.timestamp + INITIAL_DELAY);
+
+        // testing single unpause execution
+        vm.expectRevert("Unpause operation is not scheduled");
+        pauser.executeUnpauseSingle(IPausable(address(etherFiOracleInstance)), 1);
+        pauser.executeUnpauseSingle(IPausable(address(liquifierInstance)), 1);
```
seongyun-ko, line 52: line break

test/Pauser.t.sol
```diff
@@ -0,0 +1,147 @@
...
+        Pauser pauserImplementation = new Pauser();
+        initialPausables.push(IPausable(address(liquifierInstance)));
+        initialPausables.push(IPausable(address(etherFiOracleInstance)));
+        
+        bytes memory initializerData = abi.encodeWithSelector(Pauser.initialize.selector, initialPausables, INITIAL_DELAY,  address(roleRegistry));
+        pauser = Pauser(address(new UUPSProxy(address(pauserImplementation), initializerData)));
+
+        // giving the pauser contract the permissions it needs to pause and unpause the contracts
+        vm.startPrank(owner);
+        liquifierInstance.updatePauser(address(pauser), true);
```
seongyun-ko, line 25: will this be handled via the new role registry later? @solipsis

seongyun-ko (changes requested):
the fine-grained timelock proposal should enable to configure various functions to have different timelock delays, not just for pause/unpause functions. But I like the `pauser` contract as a single gateway to pause/unpause entire protocol.

I suggest:
1. (@jtfirek) let's re-write the EFIP-6 to be specifically for `pauser` contract  
2. (@solipsis ) let's create a new EFIP proposal to include @solipsis 's https://github.com/etherfi-protocol/smart-contracts/pull/84. let's add a planed configuration as well; use the role-registry system and multiple timelock contracts together

src/Pauser.sol
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * @notice Used to pause calls to `deposit` and `depositWithPermit`.
+     */
+    bool public isPaused;
+
+    /**
+     * @notice Minimum delay required to execute an unpause action.
+     */
+    uint256 public minUnpauseDelay;
```
seongyun-ko, line 29: in which scenario do we need a delay to unpause a contract?

jtfirek: I thought that it was a functional requirement of this task

jtfirek: Darn I think we had a miscommunication bc I thought this was a functional requirement of this task :(
 
Screenshot 2024-07-19 at 10 50 00 AM

jtfirek:
> the fine-grained timelock proposal should enable to configure various functions to have different timelock delays, not just for pause/unpause functions. But I like the `pauser` contract as a single gateway to pause/unpause entire protocol.
> 
> I suggest:
> 
> 1. (@jtfirek) let's re-write the EFIP-6 to be specifically for `pauser` contract
> 2. (@solipsis ) let's create a new EFIP proposal to include @solipsis 's [role-registry + granulare EFNM call whitelist #84](https://github.com/etherfi-protocol/smart-contracts/pull/84). let's add a planed configuration as well; use the role-registry system and multiple timelock contracts together

Sorry for writing a bunch of code we don't need :(. I was under the impression that we wanted timelock functionality in this contract. 
Screenshot 2024-07-19 at 10 55 10 AM

My mistake for communicating over discord instead of through efips. I think this confused us

src/Pauser.sol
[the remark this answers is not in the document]
jtfirek: Fetching directly from the role registry now.

src/Pauser.sol
jtfirek: removed in refactor

test/Pauser.t.sol
jtfirek: removed in refactor

test/Pauser.t.sol
jtfirek: added ✅

src/Pauser.sol
seongyun-ko: oh, don't feel so. I know you were working on it and thought we need that (why not, cause it is more generic). but I was imagining some use cases :)

PR #97, opened 2024-07-19
Make EtherFiNode contract EIP-1271 compatible

seongyun-ko:


solipsis (approved):
We should probably make a note to remove this after claim. So that any future message signed by any admin address is not considered valid

PR #107, opened 2024-07-23
merge master

seongyun-ko:
- extend `isValidSignature` to use the role registry
- move efip-8 to efip-11 due to conflict
- handle merge conflicts

### DIFF
--- a/src/EtherFiNode.sol
+++ b/src/EtherFiNode.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.8.24;
 
 import "./interfaces/IEtherFiNode.sol";
 import "./interfaces/IEtherFiNodesManager.sol";
-import "src/EtherFiNodesManager.sol";
 import "@openzeppelin/contracts/utils/math/Math.sol";
 import "@openzeppelin/contracts/proxy/beacon/IBeacon.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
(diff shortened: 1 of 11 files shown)


src/EtherFiNode.sol
```diff
@@ -753,6 +756,22 @@ contract EtherFiNode is IEtherFiNode {
...
+
+    /**
+     * @dev Should return whether the signature provided is valid for the provided data
+     * @param _digestHash   Hash of the data to be signed
+     * @param _signature Signature byte array associated with _data
+    */
+    function isValidSignature(bytes32 _digestHash, bytes memory _signature) public view override returns (bytes4 magicValue) {
+        bytes32 NODE_ADMIN_ROLE = keccak256("EFNM_NODE_ADMIN_ROLE");
+        (address signer, ) = ECDSA.tryRecover(_digestHash, _signature);
+        bool isAdmin = EtherFiNodesManager(payable(etherFiNodesManager)).roleRegistry().hasRole(NODE_ADMIN_ROLE, signer);
```
solipsis, line 771: should we add it to the interface and use the interface we are already importing? Not a big deal

solipsis: also potentially do we want to make this a separate role. The EFNM_NODE_ADMIN_ROLE is constrained from arbitrary actions via the whitelist. Some potentially dangerous situations with this role being able to approve arbitrary signatures

seongyun-ko: 1. good point! will do
2,3. good points. The signer then should be domain-specific. having a mapping(bytes32 -> address) would make sense?

seongyun-ko: ok, we will revisit the signing part later again during the Certora audit

PR #109, opened 2024-07-23
Remove blacklist proposal and implementation

jtfirek:
Proposal and implementation to remove the transfer blacklist specified by efip-5. If approved I will create a second pr to remove the blacklist from the `master` branch.


seongyun-ko (approved):
LGTM

PR #111, opened 2024-07-23
Remove hasRestaked() from critical paths

solipsis:
The upcoming Eigenlayer PEPE update removes the `hasRestaked()` function that our protocol uses in several locations. With the release of PEPE there will never be a state where previously hasRestaked would have been false, so we can safely remove those codepaths.

### DIFF
--- a/src/EtherFiNode.sol
+++ b/src/EtherFiNode.sol
@@ -28,7 +28,6 @@ contract EtherFiNode is IEtherFiNode, IERC1271 {
     address public eigenPod;
 
     /// @dev Is this withdrawal safe is configured for restaking within the etherfi protocol.
-    ///      Independent of whether the associated eigenpod has toggled its hasRestaked flag.
     bool public isRestakingEnabled;
 
     uint16 public version;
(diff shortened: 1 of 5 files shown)


src/EtherFiNode.sol
```diff
@@ -634,33 +626,9 @@ contract EtherFiNode is IEtherFiNode, IERC1271 {
...
-            // https://github.com/Layr-Labs/eigenlayer-contracts/tree/90a0f6aee79b4a38e1b63b32f9627f21b1162fbb/src/contracts/pods/EigenPod.sol#L685
-            // 
-            // Therefore, here we only need to queue {NonBeaconChainEthWithdrawal, full withdrawal}
-            _queueNonBeaconChainEthWithdrawal();
-            fullWithdrawalRoots = _queueRestakedFullWithdrawal();
-        }
-
-    }
-
-    function _queueNonBeaconChainEthWithdrawal() internal {
```
seongyun-ko, line 656: don't we need this anymore?

seongyun-ko (approved):
lgtm

src/EtherFiNode.sol
solipsis: both `nonBeaconChainETHBalanceWei()` and `withdrawNonBeaconChainETHBalanceWEI` are removed in PEPE. 

All eth is treated equal in the new system even if someone cheated to send eth to the pod (block rewards, self-destruct, etc.). Unlike M2 which didn't award shares for nonBeaconEth, PEPE does reward shares for nonBeaconEth

In fact in the new system, if you don't care about actual validator rewards, the cheapest way to restake with eigenlayer is to just self-destruct a bunch of ETH into the pod :)

PR #114, opened 2024-07-24
[Fix] forge fork testing

jtfirek:
When `initializeRealisticForkWithBlock(forkEnum, block.number)` is called before a fork is created, block.number defaults to 1. This causes tests calling `initializeRealisticFork()` to fail, as they expect the fork to be created at the current block.


solipsis (approved):

vvalecha519 (approved):

vvalecha519:
can you apply that change to the testnet as well

PR #116, opened 2024-07-25
clean LiquidityPool

seongyun-ko:
- Deprecate whitelist
- Combine two separate calls flows by `isLpBnftHolder`
- Remove `_depositRoot` parameter from `register` flow


solipsis (approved):
love me a red PR

PR #119, opened 2024-07-25
[Implementation][EFIP-4] Process Fees in EETH

vvalecha519:
https://hackmd.io/@sK3pnSmVSxekZ0_510u48g/rkBMangsA

### DIFF
--- a/src/LiquidityPool.sol
+++ b/src/LiquidityPool.sol
@@ -22,7 +22,6 @@ import "./interfaces/ILiquidityPool.sol";
 import "./interfaces/IEtherFiAdmin.sol";
 import "./interfaces/IAuctionManager.sol";
 import "./interfaces/ILiquifier.sol";
-import "forge-std/console.sol";
 
 contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, ILiquidityPool {
     //--------------------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


src/LiquidityPool.sol
```diff
@@ -461,6 +462,20 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, IL
 
         emit Rebase(getTotalPooledEther(), eETH.totalShares());
     }
+    /// @notice pay protocol fees including 5% to treaury, 5% to node operator and ethfund bnft holders
+    /// @param _protocolFees The amount of protocol fees to pay in ether
+    function payProtocolFees(uint128 _protocolFees) external {
+        if (msg.sender != address(etherFiAdminContract)) revert IncorrectCaller();
```
seongyun-ko, line 468: Is it necessary though the caller is being checekd on `depositToRecipient` function

seongyun-ko:
- could u add EFIP-4 proposal to it too? https://github.com/etherfi-protocol/smart-contracts/blob/staging-2.5/proposals/EFIP-4.md
- could u update EFIP-4 proposal regarding the one-time big payment processing? we will need to comm this or get prepared at least

src/LiquidityPool.sol
vvalecha519: not necessary just more clear and its a little stronger because if I remove it liquifier contract can call method.

seongyun-ko (approved):
lgtm

seongyun-ko:
can u also update EFIP-4 with the audit reviews from Nethermind (link its reference) and Certora (upload the draft review and link the reference)

PR #124, opened 2024-07-26
[EFIP-10] Whitelisted Delegate calls for EtherFiNode/EigenPod

seongyun-ko:
(made from @solipsis 's work merged to staging-2.5 for mainnet)

Currently, we use the blacklisting to restrict blacklisted functions from being executed. However, the blacklisting approach is not robust and can be bypassed by malicious actors by upgrade. Introducing the whitelisting will mitigate these risks by ensuring only allowed operations can be executed.

This proposes the introduction of a whitelist mechanism on the delegate call into the EtherFiNode/EigenPod contracts. This change aims to enhance security by restricting call forwarding and function execution to a predefined list of delegates, ensuring that only authorized operations can be performed.

### DIFF
--- a/proposals/EFIP-10.md
+++ b/proposals/EFIP-10.md
@@ -1,0 +1,34 @@
+# [EFIP-10] Whitelisted Delegate calls for EtherFiNode/EigenPod
+
+
+**Author**: dave (dave@ether.fi), syko (seongyun@ether.fi)
+
+**Date**: 2024-07-22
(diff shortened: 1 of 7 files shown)


solipsis (commented):
I am a little confused. What are the differences here between the work I did on this already and the EFIP that I am supposed to write this week?

solipsis (approved):

PR #126, opened 2024-07-26
[Chore] [EFIP-5] - acknowledge findings from harborn audit 

jtfirek:
- Cache array length outside of loop 
- Add whitelist events
- Removing redundant console import


seongyun-ko (approved):
lgtm

PR #127, opened 2024-07-26
[Chore] [EFIP-5] - Remove blacklist and ack harborn audit

jtfirek:
- Remove blacklist
- Cache array length outside of loop 
- Add whitelist events
- Removing redundant console import


jtfirek:
@seongyun-ko Shouldn't we still merge this into master so removal of the blacklist is consistent with `2.5 staging`? I thought that we wanted to get this before the certora audit was complete?

PR #128, opened 2024-07-26
[EFIP-14] Role Registry proposal

solipsis:
This PR introduces an EFIP for improving our permissions and security across the protocol by using a central Role Registry contract


seongyun-ko (approved):
lgtm

Issue #24, opened 2024-03-26
WithdrawRequestNFT's `claimWithdraw` is wrong in handling when there was fee


seongyun-ko:
<img width="668" alt="image" src="https://github.com/etherfi-protocol/smart-contracts/assets/6564103/65fa00a1-8478-466c-aca8-87179d3bf260">

https://etherscan.io/tx/0x188e077e92f10c9ca222cead4ab78e773c6ba330ecc12e4c89a71c88eeda1348

https://dashboard.tenderly.co/ether-fi/ether-fi/tx/mainnet/0x188e077e92f10c9ca222cead4ab78e773c6ba330ecc12e4c89a71c88eeda1348/state-diff

<img width="650" alt="image" src="https://github.com/etherfi-protocol/smart-contracts/assets/6564103/2c606658-e0d4-4a60-8f24-4bd58138c38f">

it is wrongly counting the unburned share due to the fee….

the correct fix would be like:

```
function claimWithdraw(uint256 tokenId) public {
    require(ownerOf(tokenId) == msg.sender, "Not the owner of the NFT");

    IWithdrawRequestNFT.WithdrawRequest memory request = _requests[tokenId];
    uint256 fee = uint256(request.feeGwei) * 1 gwei;
    uint256 amountToWithdraw = getClaimableAmount(tokenId);

    // transfer eth to requester
    address recipient = ownerOf(tokenId);
    _burn(tokenId);
    delete _requests[tokenId];

    uint256 amountBurnedShare = 0;
    if (fee > 0) {
        // send fee to membership manager
        amountBurnedShare += liquidityPool.withdraw(address(membershipManager), fee);
    }

    amountBurnedShare += liquidityPool.withdraw(recipient, amountToWithdraw);
    uint256 amountUnBurnedShare = request.shareOfEEth - amountBurnedShare;
    if (amountUnBurnedShare > 0) {
        accumulatedDustEEthShares += uint96(amountUnBurnedShare);
    }

    emit WithdrawRequestClaimed(uint32(tokenId), amountToWithdraw + fee, amountBurnedShare, recipient, fee);
}
```

so, the protocol is not calling `burnAccumulatedDustEEthShares` function.
We should fix it and upgrade the contract

seongyun-ko:
fixed

PR #135, opened 2024-08-05
fix the accounting error in dust share for withdrawal requests

seongyun-ko:
Handle https://github.com/etherfi-protocol/smart-contracts/issues/24

### DIFF
--- a/src/WithdrawRequestNFT.sol
+++ b/src/WithdrawRequestNFT.sol
@@ -114,12 +114,13 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
         _burn(tokenId);
         delete _requests[tokenId];
 
+        uint256 amountBurnedShare = 0;
         if (fee > 0) {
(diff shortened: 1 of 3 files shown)


jtfirek (approved):

jtfirek:
ltgm! My only question is can we redirect the fee to a more useful place than `membershipManager` or remove the fee all together at some point?

jtfirek (changes requested):
We should remove the fee

jtfirek (commented):
Should we go head and remove the ether.fi burn fee completely? This has been deprecated and jv has to periodically do manual refunds to users when they complain about this

seongyun-ko:
> Should we go head and remove the ether.fan burn fee completely? We have told users there is no longer a burn fee and jv has to periodically send manual refunds to users when they complain about this
> 
> I can build off this PR if you want me to implement this.

we can update the fee https://etherscan.io/address/0x3d320286E014C3e1ce99Af6d6B00f0C1D63E3000#writeProxyContract#F13

PR #136, opened 2024-08-05
allow 'WITHDRAW_NFT_ADMIN_ROLE' can perform the fast-track withdrawal

seongyun-ko:
- Currently, the protocol relies entirely on the Oracle to process the withdrawals
- this PR enables the fast-track withdrawal processing (which does not require the Oracle consensus) to be used in the emergency such as de-peg

EFIP will follow


jtfirek (commented):
I think it would be hard for a non-syko human admin to quickly derive an input for `finalizeRequests`.

Should we add a convenience function for  calculating how much `EthAmountLockedForWithdrawal` will increase for a given `lastRequestId` input?

Or a revert statement that won't allow the admin to finalize more withdrawals than can be supported by the current amount of ETH in the LP?

seongyun-ko:
makes sense! could you quickly add  a convenience function for calculating how much EthAmountLockedForWithdrawal will increase for a given lastRequestId input??

PR #137, opened 2024-08-05
[HotFix] Enable partial redemption of stETH 

seongyun-ko:
Due to the wrong branch expression in `Liquifier.stEthRequestWithdrawal(uint256 amount)`, we were not able to trigger the partial redemption request for stETH. This PR fixes it


vvalecha519 (approved):

seongyun-ko:
Will be shipped with v2.5 upgrade later

PR #139, opened 2024-08-08
Enable partial redemption of stETH, add support for swapEEthForStEth …

seongyun-ko:
- Enable partial redemption of stETH
- add support for swapEEthForStEth to swap eETH for stETH
- fix broken tests

### DIFF
--- a/src/Liquifier.sol
+++ b/src/Liquifier.sol
@@ -250,7 +250,8 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
 
     function stEthRequestWithdrawal(uint256 _amount) public returns (uint256[] memory) {
         if (!roleRegistry.hasRole(LIQUIFIER_ADMIN_ROLE, msg.sender)) revert IncorrectRole();
-        if (_amount < lidoWithdrawalQueue.MIN_STETH_WITHDRAWAL_AMOUNT() || _amount > lido.balanceOf(address(this))) revert IncorrectAmount();
+        if (_amount < lidoWithdrawalQueue.MIN_STETH_WITHDRAWAL_AMOUNT()) revert IncorrectAmount();
+        if (_amount > lido.balanceOf(address(this))) revert NotEnoughBalance();
 
         tokenInfos[address(lido)].ethAmountPendingForWithdrawals += uint128(_amount);
 
@@ -293,8 +294,7 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
     // Swap Liquifier's eETH for ETH from the liquidity pool and send it back to the liquidity pool
     function withdrawEEth(uint256 amount) external {
         if (!roleRegistry.hasRole(LIQUIFIER_ADMIN_ROLE, msg.sender)) revert IncorrectRole();
-        liquidityPool.withdraw(address(this), amount);
-        _withdrawEther();
+        liquidityPool.withdraw(address(liquidityPool), amount);
     }
 
     function updateWhitelistedToken(address _token, bool _isWhitelisted) external onlyOwner {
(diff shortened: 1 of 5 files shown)


src/Liquifier.sol
```diff
@@ -204,6 +205,15 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
         return eEthShare;
     }
 
+    // Swap user's eETH for Liquifier's stETH
+    function swapEEthForStEth(uint256 _amount) external whenNotPaused nonReentrant {
```
jtfirek, line 209: Remove this?
Screenshot 2024-08-07 at 10 48 16 PM

jtfirek, line 209: just something to consider that this function might be heavily used by MEV bots in a event where we depeg. 

Bots can buy cheap eETH from a dex and swap it for stETH one to one here! Probably a good thing as this will help us repeg.

src/Liquifier.sol
```diff
@@ -240,7 +250,7 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
 
     function stEthRequestWithdrawal(uint256 _amount) public returns (uint256[] memory) {
         if (!roleRegistry.hasRole(LIQUIFIER_ADMIN_ROLE, msg.sender)) revert IncorrectRole();
-        if (_amount < lidoWithdrawalQueue.MIN_STETH_WITHDRAWAL_AMOUNT() || _amount < lido.balanceOf(address(this))) revert NotEnoughBalance();
+        if (_amount < lidoWithdrawalQueue.MIN_STETH_WITHDRAWAL_AMOUNT() || _amount > lido.balanceOf(address(this))) revert IncorrectAmount();
```
vvalecha519, line 253: I would split this

src/Liquifier.sol
seongyun-ko: yes! should we take fee

vvalecha519, line 209: why not give permission for liquifier to burn tokens so with don't have that funky eth transfer. Is it a security risk?

src/Liquifier.sol
```diff
@@ -277,10 +287,14 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
...
+        _withdrawEther();
+    }
 
-        uint256 amountToLiquidityPool = address(this).balance;
-        (bool sent, ) = payable(address(liquidityPool)).call{value: amountToLiquidityPool, gas: 20000}("");
-        if (!sent) revert EthTransferFailed();
+    // Swap Liquifier's eETH for ETH from the liquidity pool and send it back to the liquidity pool
+    function withdrawEEth(uint256 amount) external {
+        if (!roleRegistry.hasRole(LIQUIFIER_ADMIN_ROLE, msg.sender)) revert IncorrectRole();
+        liquidityPool.withdraw(address(this), amount);
```
jtfirek, line 296: You could change this and remove the need for the sequential  `_withdrawEther` call
```
liquidityPool.withdraw(address(liquidityPool), amount);
```

jtfirek (approved):

src/Liquifier.sol
vvalecha519: No fees. I think this function can be very useful to symbiotic vault since they get a lot of eeth, but they are required to swap eeth to steth since symbiotic doesn't take eeth. With a fee they will continue to use the same method of getting eth and then minting steth which prevents a major use case which is helpful to us since our liquidity pool doesn't get drained. 

Maybe we should do a check if lido's token is worth less than our then it reverts so bots don't take advantage (like what dave recommended)

seongyun-ko: makes sense!

src/Liquifier.sol
seongyun-ko: DONE

src/Liquifier.sol
seongyun-ko: just burning the tokens will screw up the TVL accounting. This implementation is to mimic the current flow to make sure those accountings alright.

src/Liquifier.sol
seongyun-ko: Nice suggestion! Updated as suggested. Needed to update `LiquidityPool.withdraw` to handle the case when it is sending ETH to itself.

PR #143, opened 2024-08-13
Fixes broken tests and continuous integration

jtfirek:
- fixes all the remaining broken tests (2 have been commented out that @solipsis has to rework posted PEPE)
- Refactors tests to remove duplicate code
- Adds a github actions workflow to assert all tests are passing


test/EigenLayerIntegration.t.sol
```diff
@@ -100,41 +92,35 @@ contract EigenLayerIntegraitonTest is TestSetup, ProofParsing {
...
-        _upgrade_auction_manager_contract();
-        _upgrade_node_oeprator_manager_contract();
-        _upgrade_etherfi_oracle_contract();
-
         setupRoleRegistry();
     }
+    
+    // Broken by PEPE upgrades
+    // TODO: Update test post PEPE upgrade
+    // function test_fullWithdraw_812() public {
```
seongyun-ko, line 100: is there any way to keep any unit test on `fullWithdraw`?

if you are going to add a new one, we can remove it? @solipsis

jtfirek: just fyi the `fullWithdraw` function call in this test reverts with the internal  `_distributePayouts` call attempting to distribute more ETH than there is in the Node contract

.github/workflows/run-forge-tests.yaml
```diff
@@ -0,0 +1,30 @@
+name: Run Forge Tests
+
+on:
+  pull_request:
+    branches:
+      - staging-2.5
```
vvalecha519, line 6: can we do this on all branches

jtfirek: changed to run for all PRs

PR #145, opened 2024-08-15
Add Event when Finalized Requested Updated

vvalecha519:


src/WithdrawRequestNFT.sol
```diff
@@ -252,6 +253,7 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
     function _authorizeUpgrade(address newImplementation) internal override onlyOwner {}
 
     function _finalizeRequests(uint256 lastRequestId, uint128 totalAmount) internal {
+        emit UpdateFinalizedRequestId(uint32(lastRequestId));
```
solipsis, line 256: is it useful for whatever tool you are building to include the `totalAmount` in the event?

solipsis (approved):

PR #152, opened 2024-08-28
Reduce Liquifier Size

jtfirek:
- Reduces the liquifier size to 23,052 resulting in a 1,524 margin under the size limit
- Removes all of the logic and tests related to accepting assets restaked in EL
- Removes the totalDeposit calculation. It is unused on-chain and can be obtained through iterating over events if we need the value off-chain
- There is still a bunch of unused `cbETH` and `wbETH` logic, but I opted to leave as it to make this pr less invasive 

Unrelated, but when we start working on protocol V3, I would like to migrate this logic to a new contract written from scratch


seongyun-ko (approved):
Great! LGTM. can you update Certora team that we deprecated these?

PR #154, opened 2024-08-28
EFIP-9 Nethermind: Execution Layer Rewards Router 

vvalecha519:


src/EtherFiRewardsRouter.sol
```diff
@@ -40,7 +39,7 @@ contract EtherFiRewardsRouter is OwnableUpgradeable, UUPSUpgradeable  {
     }
 
     function withdrawToLiquidityPool() external {
-        if (!roleRegistry.hasRole(ETHERFI_ROUTER_ADMIN, msg.sender)) revert IncorrectRole();
+        if (msg.sender != admin) revert IncorrectRole();
```
seongyun-ko, line 42: @vvalecha519 I think we can make this function `withdrawToLiquidityPool` permission-less

PR #156, opened 2024-09-02
Certora-Fix- Add Ability to Change Treasury

vvalecha519:
-Originally we made the treasury immutable but I think it is better to have it behind a timelock but mutable so if we change our treasury we are able to do so. In terms of security it does not make a difference because treasury address is only used for recovery


jtfirek (changes requested):
You should configure your editor to newline all your files by default so you don't introduce these git warnings by accident

test/EtherFiRewardsRouter.t.sol
```diff
@@ -54,5 +65,28 @@ contract EtherFiRewardsRouterTest is TestSetup {
...
+        etherfiRewardsRouterInstance.setTreasury(alice);
+        assertEq(etherfiRewardsRouterInstance.treasury(), alice);
+    } 
+
+    function test_setTreasuryFailure() public {
+        vm.startPrank(bob);
+        vm.expectRevert("Ownable: caller is not the owner");
+        etherfiRewardsRouterInstance.setTreasury(alice);
+    } 
 }
```
jtfirek, line 92: You should configure your editor to newline all your files by default so you don't introduce these git warnings by accident

jtfirek:
Shouldn't we be using a initializer to set these values?

https://github.com/etherfi-protocol/smart-contracts/blob/84a3f60f247e941d0f8c878d43ebca7259abf1b7/src/EtherFiRewardsRouter.sol#L28C1-L32C6

test/EtherFiRewardsRouter.t.sol
vvalecha519: thx

jtfirek (approved):
ltgm

PR #158, opened 2024-09-04
[Certora][H-02] NFT Withdrawal Price Finalization

jtfirek:
### Motivation
This pr addresses the concerns with the withdrawal flow raised by the certora team during the audit:
- The withdrawal requests claim process is unfair as users are exposed to the ramifications of a negative rebase after their request is finalized
- eETH requested for withdrawal still accrue staking rewards after requests are made. These positive rebase rewards remain in this contract after users claim their withdrawal, but there is currently no accounting or method to withdraw these funds as revenue

### Features
- A new mechanism to finalize the value of request when its corresponding batch of withdrawals is processed by the oracle. Note that users are still exposed to negative rebases that occur *before*  finalization
- Logic to query and claim the excess staking rewards mentioned above
- Removal of unnecessary accounting like `EthAmountLockedForWithdrawal` in the new flow
- Refactor the codebase to consistently use uint32 for withdrawal NFT id values  

### Migration
- To make the existing unclaimed finalized withdrawals work with the new system, `lastFinalizedRequestId` will be reset to 0 and the oracle will re-finalize all withdrawals with the new logic. It will be much to expensive to call `calculateTotalPendingAmount` on 30,000ish withdrawals. We will have to calculate the amount needed to re-finalized all the withdrawals off-chain and pass it on-chain with `finalizeRequests(uint256 lastRequestId, uint256 totalAmount)`. The ability to withdrawal will be temporary paused to maintain correct accounting

### DIFF
--- a/src/WithdrawRequestNFT.sol
+++ b/src/WithdrawRequestNFT.sol
@@ -23,7 +23,7 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
 
     uint32 public nextRequestId;
     uint32 public lastFinalizedRequestId;
-    uint96 public accumulatedDustEEthShares; // to be burned or used to cover the validator churn cost
+    uint96 public DEPRECATED_accumulatedDustEEthShares; // to be burned or used to cover the validator churn cost
 
     FinalizationCheckpoint[] public finalizationCheckpoints;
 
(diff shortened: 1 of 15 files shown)


jtfirek:
fyi, The failing unit tests on the last run are from Eigenpod changes from the PEPE upgrade

src/WithdrawRequestNFT.sol
```diff
@@ -23,21 +27,43 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
 
     uint32 public nextRequestId;
     uint32 public lastFinalizedRequestId;
-    uint96 public accumulatedDustEEthShares; // to be burned or used to cover the validator churn cost
+    uint96 public DEPRECATED_accumulatedDustEEthShares; // to be burned or used to cover the validator churn cost
+
+    /// Stores the cached share price at the time of finalization for each batch
+    /// For a checkpoint index i, this batch includes requests:
+    /// `finalizationCheckpoints[i-1].lastFinalizedRequestId < requestId <= finalizationCheckpoints[i].lastFinalizedRequestId`
+    FinalizationCheckpoint[] public finalizationCheckpoints;
```
vvalecha519, line 35: I think you can get rid of binary search and just use a map of lastFinalizedRequestId -> finalizedPrice 
and add lastFinalizedRequestId in WithdrawRequest

jtfirek: - The problem with this approach is that the oracle would have to update the lastFinalizedRequestId for each withdrawal request which is too expensive of an operational cost
- When using the binary search the oracle only has to push 1 element onto the `finalizationCheckpoints` vs having to update the `WithdrawRequest` for each request being finalized

PR #161, opened 2024-09-08
Deposit Adapter Contract

jtfirek:
- Allows for atomic swaps to weETH from {ETH, wETH, stETH, wstETH} 
- Original plan was to provide these features in a separate contracts without additions to our core contracts, but sadly this approach incorrectly populates the sender field of the deposit event which the growth team relies on. The functions added to liquifier and liquidity pool fix this

### DIFF
--- a/script/deploys/DeployEtherFISuite.s.sol
+++ b/script/deploys/DeployEtherFISuite.s.sol
@@ -165,7 +165,7 @@ contract DeployEtherFiSuiteScript is Script {
         wstETH = new TestERC20("Coinbase ETH", "cbEth");
         sfrxEth = new TestERC20("Frax ETH", "sfrxEth");
 
-        liquidityPoolImplementation = new LiquidityPool(address(0));
+        liquidityPoolImplementation = new LiquidityPool();
         liquidityPoolProxy = new UUPSProxy(
             address(liquidityPoolImplementation),
             ""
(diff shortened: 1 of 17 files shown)


solipsis:
If the blocker on your ideal solution is the transfer event, would it be better for us to just make a new event with with the sender field the growth team expects and they can just swap to the new event?

jtfirek:
> If the blocker on your ideal solution is the transfer event, would it be better for us to just make a new event with with the sender field the growth team expects and they can just swap to the new event?

This is a good point! The points printer is already janky so this definitely a better solution that adding weird code to our core contracts

jtfirek:
I updated the code using dave's suggestion. Now requires 0 changes to our core contracts!

seongyun-ko (approved):
LGTM

Shivam78288:
You should check the amount of stETH that actually got transferred to the contract by taking the delta of before and after. A lot of times, the stETH amount transferred is 1 wei lesser and it may cause the tx to revert.

jtfirek:
> You should check the amount of stETH that actually got transferred to the contract by taking the delta of before and after. A lot of times, the stETH amount transferred is 1 wei lesser and it may cause the tx to revert.

Yea I had to use `assertApproxEqAbs` quite a bit due to output amounts being 1-2 wei less than expected, but it doesn't seem to result in reverts due to the fact that stETH `transferFrom` is transferring shares under the hood.

I added an additional delta checks, but lmk if you were looking for something more specific!

seongyun-ko:
An extension would be to allow deposit of ETH for {weETHs, weETHk}

PR #180, opened 2024-10-03
[Certora] [M-10] + [M-11] Manual tracking of EETH shares owned by membership manager

solipsis:
This PR adds a manual tracking for the number of EETH shares owned by the membership manager. This prevents anyone from being able to impact rebasing operations via donating eETH shares to the membership manager

Primary testing provided by the existing `test_bring_random_monkeys()` function. At the conclusion of this function with thousands of deposit / rebase / withdrawal operations, shares are within 5000 shares of the target amount which is just dust.

### DIFF
--- a/src/MembershipManager.sol
+++ b/src/MembershipManager.sol
@@ -65,6 +65,9 @@ contract MembershipManager is Initializable, OwnableUpgradeable, PausableUpgrade
 
     IEtherFiAdmin public etherFiAdmin;
 
+    // Phase 2.5
+    uint256 public membershipShares;
+
     //--------------------------------------------------------------------------------------
     //-------------------------------------  EVENTS  ---------------------------------------
     //--------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/MembershipManager.sol
```diff
@@ -167,10 +176,12 @@ contract MembershipManager is Initializable, OwnableUpgradeable, PausableUpgrade
         (uint256 totalBalance, uint256 feeAmount) = _withdrawAndBurn(_tokenId);
 
         // transfer 'eEthShares' of eETH to the owner
+        membershipShares -= liquidityPool.sharesForAmount(totalBalance - feeAmount);
         eETH.transfer(msg.sender, totalBalance - feeAmount);
 
         if (feeAmount > 0) {
-            liquidityPool.withdraw(address(this), feeAmount);
+            uint256 feeShares = liquidityPool.withdraw(address(this), feeAmount);
```
seongyun-ko, line 183: By this `withdraw`, the membership manager contract will receive ETH (not eETH), which is used to boost the MembershipNFT APR when rebasing happens. so it should not be counted as `membershipShares`

* ETH balance in MembershipManager https://etherscan.io/address/0x3d320286E014C3e1ce99Af6d6B00f0C1D63E3000

seongyun-ko (changes requested):
LGTM overall, but left a feedback on one fix

A few points to confirm:
- is the contract size below the limit?
- isnt' there any change required here? https://github.com/etherfi-protocol/smart-contracts/blob/audit-m10-m11/src/libraries/GlobalIndexLibrary.sol

PR #185, opened 2024-10-11
[Certora][I-03] fix double role check

solipsis:
remove onlyOwner modifier since it is already scoped to a particular role


jtfirek (commented):
lgtm

PR #186, opened 2024-10-15
Whitelist EigenLayer RewardsCoordinator's processClaim to claim AVS rewards

seongyun-ko:
In order to enable rewards claiming via call forwarding I need you to please whitelist function `0x3ccc861d` on contract `0x7750d328b314EfFa365A0402CcfD489B80B0adda`

which is processClaim on eigenlayers mainnet RewardsCoordinator contract. Please note that this needs to be added to the external calls, not the eigenpod calls because the earner is the pod owner (etherfi node) not the eigenpod.

```
    function updateAllowedForwardedExternalCalls(bytes4 _selector, address _target, bool _allowed) external onlyAdmin {
```


solipsis (approved):

jtfirek (approved):
Transaction matches the one included in the notion doc

PR #188, opened 2024-10-16
Liquifier: pricing of stETH via CurvePool & fixed rate fee

seongyun-ko:
## Context

The arbitrageurs are exploiting ether.fi's fast withdrawal processing to redeem their stETH faster.
+ some bugs

## What this PR is about
- use CurvePool to price the stETH
- apply the fied deposit fee rate
- enable partial redemption of stETH
- EigenLayer's queued withdrawal struct was refactored from [this](https://github.com/code-423n4/2023-04-eigenlayer/blob/main/src/contracts/core/StrategyManager.sol#L400-L407) to [this](https://github.com/Layr-Labs/eigenlayer-contracts/blob/master/src/contracts/core/DelegationManager.sol#L361-L369). Then, EL team has migrated manually through [it](https://github.com/Layr-Labs/eigenlayer-contracts/blob/master/src/contracts/core/DelegationManager.sol#L349C39-L349C55). Now, our old mapping keeping the old versions of `withdrawalRoots` does not work anymore. so I am removing that sanity check to process the completed withdrawals.


jtfirek:
Our FE currently displays a hard coded 1 to 1 exchange for stETH -> eETH. We could transition to using `quoteByMarketValue` to get the rate, but this value won't be accurate if the fee has been applied. 
This pr includes a function our FE can query to display the proper rate to our users:
https://github.com/etherfi-protocol/smart-contracts/pull/189

jtfirek (approved):

shivam-ef (approved):
LGTM

PR #189, opened 2024-10-16
Discount Pricing Function

jtfirek:
Provides a function our FE can query to get the amount of eETH a user will receive when the discount rate is applied


seongyun-ko (approved):
lgtm

PR #203, opened 2024-11-20
v2.5 Protocol Upgrade

vvalecha519:


seongyun-ko:
upgrade v2.5 is refactored to the reduced scope with additional features in v2.49 upgrade. @vvalecha519

PR #204, opened 2024-11-20
update master with deployed contracts

vvalecha519:


src/EtherFiRewardsRouter.sol
```diff
@@ -0,0 +1,73 @@
...
+import "./RoleRegistry.sol";
+
+contract EtherFiRewardsRouter is OwnableUpgradeable, UUPSUpgradeable  {
+    using SafeERC20 for IERC20;
+
+    address public immutable treasury;
+    address public immutable liquidityPool;
+    RoleRegistry public immutable roleRegistry;
+
+    bytes32 public constant ETHERFI_ROUTER_ADMIN = keccak256("ETHERFI_ROUTER_ADMIN");
```
jtfirek, line 18: would like to note that we should upgrade this contract to remove this role in the future

jtfirek (approved):

PR #102, opened 2024-07-22
[EFIP-5] {Permit Whitelist, Transfer Blacklist}

seongyun-ko:

PR #206, opened 2024-12-04
Revert "[EFIP-5] {Permit Whitelist, Transfer Blacklist}"

seongyun-ko:
Reverts etherfi-protocol/smart-contracts#102

We didnt decide it to ship it to mainnet, yet.
thus reverting


jtfirek (approved):
lgtm, we could also remove the proposal doc as well since the feature was never shipped:
https://github.com/etherfi-protocol/smart-contracts/blob/master/proposals/EFIP-5.md

seongyun-ko:
> lgtm, we could also remove the proposal doc as well since the feature was never shipped: https://github.com/etherfi-protocol/smart-contracts/blob/master/proposals/EFIP-5.md

we can leave the proposal since it is proposal with good intentions :)

PR #207, opened 2024-12-09
weETH withdrawal: Instant withdrawal with Fee + Implicit withdrawal fee handling

seongyun-ko:
### Features

1.  Instant Withdrawal with Fee
    - Contract: EtherFiWithdrawalBuffer
    - (1) rate limits, (2) low-water mark mechanism, (3) fee & split configuration
 
2. Handling of the unburned shares during the delayed withdrawal 
   - Contract: WithdrawRequestNFT
   - the existing delayed withdrawal system had a bug in tracking the dust share amount.
   - This PR (1) fixes the bug, (2) adds the function to handle the accumulated dust, and (3) adds the fee split logic

Audited by Certora


src/EtherFiWithdrawalBuffer.sol
```diff
@@ -0,0 +1,249 @@
...
+     * @param eEthAmount The amount of eETH to redeem after the exit fee.
+     * @param receiver The address to receive the redeemed ETH.
+     * @param owner The address of the owner of the eETH.
+     * @return The amount of ETH sent to the receiver and the exit fee amount.
+     */
+    function redeemEEth(uint256 eEthAmount, address receiver, address owner) public whenNotPaused nonReentrant returns (uint256, uint256) {
+        require(canRedeem(eEthAmount), "EtherFiWithdrawalBuffer: Exceeded total redeemable amount");
+        require(eEthAmount <= eEth.balanceOf(owner), "EtherFiWithdrawalBuffer: Insufficient balance");
+
+        uint256 beforeEEthAmount = eEth.balanceOf(address(this));
```
jtfirek, line 82: What is the purpose of `beforeEEthAmount` and `afterEEthAmount`? Wouldn't just this do the same thing:
```
IERC20(address(eEth)).safeTransferFrom(owner, address(this), eEthAmount);
return _redeem(eEthAmount, receiver);
```

seongyun-ko: yes, since eETH is rebasable token and `transfer` creates different corner cases (1-wei off error)

src/WithdrawRequestNFT.sol
```diff
@@ -22,16 +28,19 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
 
     uint32 public nextRequestId;
     uint32 public lastFinalizedRequestId;
-    uint96 public accumulatedDustEEthShares; // to be burned or used to cover the validator churn cost
+    uint16 public shareRemainderSplitToTreasuryInBps;
```
jtfirek, line 31: We are changing the use of a storage variable here without an initialization function. Just need to make sure we call `updateShareRemainderSplitToTreasuryInBps` upon upgrade

src/WithdrawRequestNFT.sol
```diff
@@ -120,13 +126,33 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
...
-    function burnAccumulatedDustEEthShares() external onlyAdmin {
-        require(eETH.totalShares() > accumulatedDustEEthShares, "Inappropriate burn");
-        uint256 amount = accumulatedDustEEthShares;
-        accumulatedDustEEthShares = 0;
+    // There have been errors tracking `accumulatedDustEEthShares` in the past.
+    // - https://github.com/etherfi-protocol/smart-contracts/issues/24
+    // This is a one-time function to handle the remainder of the eEth shares after the claim of the withdraw requests
+    // It must be called only once with ALL the requests that have not been claimed yet.
+    // there are <3000 such requests and the total gas spending is expected to be ~9.0 M gas.
+    function handleAccumulatedShareRemainder(uint256[] memory _reqIds) external onlyOwner {
```
jtfirek, line 134: Could use `calldata` instead of `memory` for 1 percent gas efficiency gain

src/EtherFiWithdrawalBuffer.sol
jtfirek: 👍

src/WithdrawRequestNFT.sol
jtfirek, line 134: I think this method will likely fail given that it must be called via timelock:

- timelock transaction is scheduled with valid `_reqIds` array
- users claim withdrawals from the `_reqIds` array during 3 day timelock enforced delay
- timelock transaction then reverts when we go to execute it

See POC of failure:
https://github.com/etherfi-protocol/smart-contracts/pull/214/files

seongyun-ko: resolved!

shivam-ef (approved):
LGTM

src/WithdrawRequestNFT.sol
```diff
@@ -93,63 +127,48 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
...
     }
 
-    // a function to transfer accumulated shares to admin
-    function burnAccumulatedDustEEthShares() external onlyAdmin {
-        require(eETH.totalShares() > accumulatedDustEEthShares, "Inappropriate burn");
-        uint256 amount = accumulatedDustEEthShares;
-        accumulatedDustEEthShares = 0;
+    // This function is used to aggregate the sum of the eEth shares of the requests that have not been claimed yet.
+    // To be triggered during the upgrade to the new version of the contract.
+    function aggregateSumEEthShareAmount(uint256 _numReqsToScan) external {
```
jtfirek, line 153: Remainder shares will still accumulate in the the contract after this upgrade for future withdrawal requests due to the known diff between current share price and share price at time of request. Because of this, the admin will have to continuously call `aggregateSumEEthShareAmount` and `handleRemainder`.

The current implementation can only handle remainders for requests made before the upgrade as `_lastRequestIdToScanUntilForShareRemainder` is only set once in the initialization function:
```
_lastRequestIdToScanUntilForShareRemainder = nextRequestId - 1;
```

src/EtherFiWithdrawalBuffer.sol
```diff
@@ -0,0 +1,280 @@
+// SPDX-License-Identifier: MIT
```
jtfirek, line 1: duplicate of `EtherFiRedemptionManager`

src/WithdrawRequestNFT.sol
```diff
@@ -48,18 +70,30 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
         nextRequestId = 1;
     }
 
+    function initializeOnUpgrade(address _pauser) external onlyOwner {
+        require(pauser == address(0), "Already initialized");
+
```
jtfirek, line 75: Initialize `shareRemainderSplitToTreasuryInBps` here for best practice?

src/WithdrawRequestNFT.sol
```diff
@@ -173,18 +192,13 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
...
-            liquidityPool.reduceEthAmountLockedForWithdrawal(uint128(ethAmount));
-        }
-
         _requests[requestId].isValid = false;
 
         emit WithdrawRequestInvalidated(uint32(requestId));
     }
 
     function validateRequest(uint256 requestId) external onlyAdmin {
+        require(_exists(requestId), "Request does not exist22");
```
jtfirek, line 201: Some 2s slipped into the requires statement

test/WithdrawRequestNFT.t.sol
```diff
@@ -377,45 +319,244 @@ contract WithdrawRequestNFTTest is TestSetup {
...
-        uint256 claimableAmount = withdrawRequestNFTInstance.getRequest(requestId).amountOfEEth;
+    function test_handleRemainder() public {
+        test_aggregateSumEEthShareAmount();
 
-        // REVERT if not owner
-        vm.prank(alice);
-        vm.expectRevert("Ownable: caller is not the owner");
-        withdrawRequestNFTInstance.seizeInvalidRequest(requestId, chad);
+        vm.startPrank(withdrawRequestNFTInstance.owner());
+        vm.expectRevert("Not all requests have been scanned");
```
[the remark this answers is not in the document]
jtfirek:
```
[FAIL: Error != expected error: Not all prev requests have been scanned != Not all requests have been scanned] test_handleRemainder() (gas: 5437442)
```

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,280 @@
...
+        uint256 ethReceived = address(this).balance - prevBalance;
+
+        // To Stakers by burning shares
+        eEth.burnShares(address(this), feeShareToStakers);
+
+        // To Treasury by transferring eETH
+        IERC20(address(eEth)).safeTransfer(treasury, eEthFeeAmountToTreasury);
+        
+        // To Receiver by transferring ETH
+        (bool success, ) = receiver.call{value: ethReceived, gas: 100_000}("");
```
jtfirek, line 151: given its just an ETH transfer, it is fine to not specify a gas limit here like we do in the LiquidityPool Contract:
```
function _sendFund(address _recipient, uint256 _amount) internal {
        uint256 balanace = address(this).balance;
        (bool sent, ) = _recipient.call{value: _amount}("");
        require(sent && address(this).balance == balanace - _amount, "SendFail");
    }
```
```suggestion
        (bool success, ) = receiver.call{value: ethReceived}("");

Comment (jtfirek) [reply src/EtherFiRedemptionManager.sol:151]: If we have a reason to keep it, we could lower it to 50k and still be safe for basic operations in the receiver fallback

Comment (seongyun-ko) [reply src/EtherFiRedemptionManager.sol:151]: thanks. reduced to 10k

**Review by jtfirek — APPROVED** [review]

File: src/EtherFiRedemptionManager.sol [hunk 10]
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    function _convertFromBucketUnit(uint64 bucketUnit) internal pure returns (uint256) {
+        return bucketUnit * BUCKET_UNIT_SCALE;
+    }
+
+
+    function _calcRedemption(uint256 ethAmount) internal view returns (uint256 eEthShares, uint256 eEthAmountToReceiver, uint256 eEthFeeAmountToTreasury, uint256 sharesToBurn, uint256 feeShareToTreasury) {
+        eEthShares = liquidityPool.sharesForAmount(ethAmount);
+        eEthAmountToReceiver = liquidityPool.amountForShare(eEthShares.mulDiv(BASIS_POINT_SCALE - exitFeeInBps, BASIS_POINT_SCALE)); // ethShareToReceiver
```
shivam-ef, line 265: wouldn't this be just `ethAmount.mulDiv(BASIS_POINT_SCALE - exitFeeInBps, BASIS_POINT_SCALE)`
It would cost more gas to just call liquidity pool again

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+        uint256 eEthShareFee = eEthShares - sharesToBurn;
+        uint256 feeShareToStakers = eEthShareFee - feeShareToTreasury;
+
+        // Snapshot balances & shares for sanity check at the end
+        uint256 prevBalance = address(this).balance;
+        uint256 prevLpBalance = address(liquidityPool).balance;
+        uint256 totalEEthShare = eEth.totalShares();
+
+        // Withdraw ETH from the liquidity pool
+        assert (liquidityPool.withdraw(address(this), eEthAmountToReceiver) == sharesToBurn);
```
shivam-ef, line 145: Better to do a require statement for clear error messages

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+        uint256 ethReceived = address(this).balance - prevBalance;
+
+        // To Stakers by burning shares
+        eEth.burnShares(address(this), feeShareToStakers);
+
+        // To Treasury by transferring eETH
+        IERC20(address(eEth)).safeTransfer(treasury, eEthFeeAmountToTreasury);
+        
+        require(eEth.totalShares() >= 1 gwei && eEth.totalShares() == totalEEthShare - (sharesToBurn + feeShareToStakers), "EtherFiRedemptionManager: Invalid total shares");
```
shivam-ef, line 154: better to cache totalShares instead of calling twice for efficiency?

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+        uint256 prevBalance = address(this).balance;
+        uint256 prevLpBalance = address(liquidityPool).balance;
+        uint256 totalEEthShare = eEth.totalShares();
+
+        // Withdraw ETH from the liquidity pool
+        assert (liquidityPool.withdraw(address(this), eEthAmountToReceiver) == sharesToBurn);
+        uint256 ethReceived = address(this).balance - prevBalance;
+
+        // To Stakers by burning shares
+        eEth.burnShares(address(this), feeShareToStakers);
```
shivam-ef, line 149: We need to whitelist redemption manager as well for being able to call burn shares in liquidity pool with latest fix

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+
+        // To Treasury by transferring eETH
+        IERC20(address(eEth)).safeTransfer(treasury, eEthFeeAmountToTreasury);
+        
+        require(eEth.totalShares() >= 1 gwei && eEth.totalShares() == totalEEthShare - (sharesToBurn + feeShareToStakers), "EtherFiRedemptionManager: Invalid total shares");
+
+        // To Receiver by transferring ETH
+        (bool success, ) = receiver.call{value: ethReceived, gas: 10_000}("");
```
shivam-ef, line 157: How does this succeed? 21k gas is required for ETH transfers as far as I remember

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+        require(success, "EtherFiRedemptionManager: Transfer failed");
+
+        // Make sure the liquidity pool balance is correct && total shares are correct
+        require(address(liquidityPool).balance == prevLpBalance - ethReceived, "EtherFiRedemptionManager: Invalid liquidity pool balance");
+        require(eEth.totalShares() >= 1 gwei && eEth.totalShares() == totalEEthShare - (sharesToBurn + feeShareToStakers), "EtherFiRedemptionManager: Invalid total shares");
```
shivam-ef, line 162: Duplicate check

src/WithdrawRequestNFT.sol
```diff
@@ -196,7 +229,61 @@ contract WithdrawRequestNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgrad
...
+
+        uint256 eEthAmountToTreasury = _eEthAmount.mulDiv(shareRemainderSplitToTreasuryInBps, BASIS_POINT_SCALE);
+        uint256 eEthAmountToBurn = _eEthAmount - eEthAmountToTreasury;
+        uint256 eEthSharesToBurn = liquidityPool.sharesForAmount(eEthAmountToBurn);
+        uint256 eEthSharesToMoved = eEthSharesToBurn + liquidityPool.sharesForAmount(eEthAmountToTreasury);
+
+        totalRemainderEEthShares -= eEthSharesToMoved;
+
+        eETH.transfer(treasury, eEthAmountToTreasury);
+        eETH.burnShares(address(this), eEthSharesToBurn);
```
shivam-ef, line 269: This also needs to change because of the recent upgrade on eETH contract

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+        _redeem(eEthAmount, eEthShares, receiver, eEthAmountToReceiver, eEthFeeAmountToTreasury, sharesToBurn, feeShareToTreasury);
+    }
+
+    function redeemEEthWithPermit(uint256 eEthAmount, address receiver, IeETH.PermitInput calldata permit) external {
+        try eEth.permit(msg.sender, address(this), permit.value, permit.deadline, permit.v, permit.r, permit.s) {} catch {}
+        redeemEEth(eEthAmount, receiver);
+    }
+
+    function redeemWeEthWithPermit(uint256 weEthAmount, address receiver, IWeETH.PermitInput calldata permit) external {
+        try weEth.permit(msg.sender, address(this), permit.value, permit.deadline, permit.v, permit.r, permit.s)  {} catch {}
```
shivam-ef, line 123: msg.sender should be replaced with a new variable which should be owner otherwise no point of permit if msg sender themselves are calling this function

src/EtherFiRedemptionManager.sol
```diff
@@ -0,0 +1,301 @@
...
+        (uint256 eEthShares, uint256 eEthAmountToReceiver, uint256 eEthFeeAmountToTreasury, uint256 sharesToBurn, uint256 feeShareToTreasury) = _calcRedemption(eEthAmount);
+
+        IERC20(address(weEth)).safeTransferFrom(msg.sender, address(this), weEthAmount);
+        weEth.unwrap(weEthAmount);
+
+        _redeem(eEthAmount, eEthShares, receiver, eEthAmountToReceiver, eEthFeeAmountToTreasury, sharesToBurn, feeShareToTreasury);
+    }
+
+    function redeemEEthWithPermit(uint256 eEthAmount, address receiver, IeETH.PermitInput calldata permit) external {
+        try eEth.permit(msg.sender, address(this), permit.value, permit.deadline, permit.v, permit.r, permit.s) {} catch {}
```
shivam-ef, line 118: msg.sender should be replaced with a new variable which should be owner otherwise no point of permit if msg sender themselves are calling this function

src/EtherFiRedemptionManager.sol
seongyun-ko: it would be off by 1 wei due to the rounding. wanted get the exact number for the exact share

src/EtherFiRedemptionManager.sol
seongyun-ko: sg

src/EtherFiRedemptionManager.sol
seongyun-ko: wanted to make sure that this still holds AFTER the asset transfer. so using Cached data wont help. wdyt?

src/EtherFiRedemptionManager.sol
seongyun-ko: yes

src/EtherFiRedemptionManager.sol
seongyun-ko: 21k gas fee is the base fee per txn; not per `transfer` operation within the txn

src/EtherFiRedemptionManager.sol
seongyun-ko: this is intended to make sure the outcome is as expected. wdyt?

src/WithdrawRequestNFT.sol
seongyun-ko: yes

src/EtherFiRedemptionManager.sol
seongyun-ko: haha true. plz fix

src/EtherFiRedemptionManager.sol
seongyun-ko: same here

src/EtherFiRedemptionManager.sol
shivam-ef: makes sense

src/EtherFiRedemptionManager.sol
shivam-ef: totalShares are called in this same require line twice, I think caching can reduce some gas here but not a major optimization

src/EtherFiRedemptionManager.sol
shivam-ef: I guess its better to just add a burnShares function on Liquidity pool which can be called by this contract instead, since LP is aware of this contract already while EETH would need to add another var to track this contract addr

src/EtherFiRedemptionManager.sol
shivam-ef: Ahh yes, I got confused between 2300gas for ether transfer and 21k for a tx

src/EtherFiRedemptionManager.sol
shivam-ef: I dont think anything can change here tbh, since there is no EETH/LP interaction between the two checks

src/EtherFiRedemptionManager.sol
shivam-ef: will do

src/EtherFiRedemptionManager.sol
shivam-ef: will fix

PR #210, opened 2024-12-12
Upgrade Liquifier so that Admin can {unPause, update the deposit cap} + Deterministic Deploy via create2

seongyun-ko:
# Txn
https://app.safe.global/transactions/tx?safe=eth:0xcdd57D11476c22d265722F68390b036f3DA48c21&id=multisig_0xcdd57D11476c22d265722F68390b036f3DA48c21_0x9f2c32321dec254540d776b0f0bb96ad7bd01e816ae519aa9f4967a059e30b98

# Context
(1) Update Liquifier so that Admin can {unPause, update the deposit c…ap}, 
(2) Add Deterministic contract deployment by Create2. the factory using OZ's library is deployed at `0x6521991A0BC180a5df7F42b27F4eE8f3B192BA62` , 
(3) Improve the review process which requires the contract code verification

Now, we can verify if the contract's bytecode matches with the local contract code


script/deploys/DeployImplementationContract.s.sol
solipsis, line 1: In the general case with our upgradeable contracts, aren't we more interested in getting the proxy address at a specific address rather than the implementation?

jtfirek:
Just found a minor issue:
https://github.com/etherfi-protocol/smart-contracts/pull/211

Everything else looks as expected

foundry.toml
```diff
@@ -7,6 +7,7 @@ gas_reports = ["*"]
 optimizer_runs = 2000
 extra_output = ["storageLayout"]
 solc-version = '0.8.24'
+bytecode_hash = 'none'
```
solipsis, line 10: Have we tested whether this change possibly breaks etherscan verification? I thought etherscan used this metadata hash as part of its contract verification process

script/Create2Factory.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/Create2.sol";
+
+contract Create2Factory {
+    event Deployed(address addr, address deployer, bytes32 bytecode_hash, bytes32 salt);
```
solipsis, line 7: Do we want indexes on this event so  can easily fetch in future?

foundry.toml
seongyun-ko: Yeah ethesxan was able to verify the contract. The impl contract was deployed with the setup

script/Create2Factory.sol
seongyun-ko: Good point! I missed! i can add it on next deployment! Thanks

solipsis (approved):

PR #213, opened 2024-12-19
add EtherFiOperationParameters

seongyun-ko:
image

# Context
This contract acts as a configurable coordination layer that lets administrators update operational parameters using tags and keys. Tags function as namespaces containing key-value pairs for different settings. Admins can be granted or revoked access to update parameters within specific tags, with all changes tracked through emitted events.

The contract provides on-chain storage for protocol settings and permissions, managing both admin access (tagAdmins) and parameter values (tagKeyValues). Frontend and backend systems can then operate automatically based on these configurations.

This is a pure parameter registry, for now. but it can have additional smart contract logics and permissions depending on our needs

# Contract
Deploed at https://etherscan.io/address/0xd0ff8996db4bdb46870b7e833b7532f484fead1a

### DIFF
--- a/src/helpers/EtherFiOperationParameters.sol
+++ b/src/helpers/EtherFiOperationParameters.sol
@@ -16,18 +16,18 @@ contract EtherFiOperationParameters is UUPSUpgradeable, OwnableUpgradeable {
         _disableInitializers();
     }
 
-    function initialize() public initializer {
+    function initialize() external initializer {
         __Ownable_init();
         __UUPSUpgradeable_init();
     }
 
-    function updateTagAdmin(string memory tag, address admin, bool allowed) public {
+    function updateTagAdmin(string memory tag, address admin, bool allowed) external onlyOwner {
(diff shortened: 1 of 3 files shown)


src/helpers/EtherFiOperationParameters.sol
```diff
@@ -0,0 +1,42 @@
...
+        __UUPSUpgradeable_init();
+    }
+
+    function updateTagAdmin(string memory tag, address admin, bool allowed) public {
+        tagAdmins[tag][admin] = allowed;
+     
+        emit UpdatedAdmin(tag, admin, allowed);
+    }
+
+    function updateTagKeyValue(string memory tag, string memory key, string memory value) public {
```
vvalecha519, line 30: Why is nothing permissioned. Anyone can update tag key value...

seongyun-ko: fixed sir

vvalecha519 (approved):

PR #215, opened 2024-12-19
Set claimer for stETH restaking rewards

jtfirek:
- `EtherFiRestaker.sol` doesn't have a way to claim AVS rewards. This allows us to set a claimer for contract's rewards


seongyun-ko (approved):
lgtm!

PR #217, opened 2024-12-24
[EigenLayer Slashing]

seongyun-ko:
By @solipsis 

This PR makes several small updates to prepare for eigenlayers upcoming slashing release. A majority of the changes in the current iteration are simply updates to our locally vendored version of several eigenlayer interfaces and helper libraries (They were previously vendored directly instead of through a submodule due to several eigenlayer files relying on an incompatible compiler version to our core contracts, they have changed that this release though so we can possibly move to vendoring them via a submodule like our other dependencies)

Changes actually effecting protocol logic are as follows:
- several withdrawal functions no longer take a middlewareTimesIndexes parameter
- natively restaked eth that is slashed will be locked in eigenpod instead of being directly burned, as a result we cannot use address(eigenpod).balance in any calculations of TVL anymore
- Updates to the EtherfiRestaker contract to use the new withdrawal functions and updated share calculations

More details about eigenlayer's changes can be found here https://github.com/eigenfoundation/ELIPs/blob/main/ELIPs/ELIP-002.md


solipsis (approved):

solipsis:
This PR makes several small updates to prepare for eigenlayers upcoming slashing release. A majority of the changes in the current iteration are simply updates to our locally vendored version of several eigenlayer interfaces and helper libraries (They were previously vendored directly instead of through a submodule due to several eigenlayer files relying on an incompatible compiler version to our core contracts, they have changed that this release though so we can possibly move to vendoring them via a submodule like our other dependencies)

Changes actually effecting protocol logic are as follows:
* several withdrawal functions no longer take a `middlewareTimesIndexes` parameter
* natively restaked eth that is slashed will be locked in eigenpod instead of being directly burned, as a result we cannot use `address(eigenpod).balance` in any calculations of TVL anymore
* Updates to the `EtherfiRestaker` contract to use the new withdrawal functions and updated share calculations

More details about eigenlayer's changes can be found here https://github.com/eigenfoundation/ELIPs/blob/main/ELIPs/ELIP-002.md

PR #220, opened 2024-12-27
Super Optimized Auction Manager

shivam-ef:
# Gas Optimization: Smarter Bid Management in Auction System

## Overview
This PR introduces a more efficient way to handle bids in our auction system. By grouping bids together and using a clever tracking system based on bit manipulation, we significantly reduce gas costs for our users while maintaining all functionality, thanks to @solipsis for the great optimization technique.

## Why This Matters
- Lower gas costs for users creating and managing bids
- More efficient contract operations
- Better scalability for the platform
- Maintains same user experience with better performance

## Impact on Users
- Lower transaction costs for bid creation 
- No changes to user interface or experience
- All existing bids continue to work as before

### DIFF
--- a/src/AuctionManager.sol
+++ b/src/AuctionManager.sol
@@ -365,4 +365,9 @@ contract AuctionManager is
         require(admins[msg.sender], "Caller is not the admin");
         _;
     }
+
+    function updateSelectedBidsInformation(
+        uint256 _batchBidId,
+        uint216 _bidAcceptBitmap
+    ) external returns (uint256[] memory) {}
 }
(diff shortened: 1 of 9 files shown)


seongyun-ko (changes requested):
can u modify the auction manager contract in-place so that we can see the diff for the upgrade

solipsis:
it looks like the popcount library is unused since it looks like you decided not to implement the bulk bid accepting. (which is fine, I know bulk accepting requires more changes to the interface). I just want to make sure that matches your expectations

shivam-ef:
> it looks like the popcount library is unused since it looks like you decided not to implement the bulk bid accepting. (which is fine, I know bulk accepting requires more changes to the interface). I just want to make sure that matches your expectations

Ohh I must have missed it, will take a look today

seongyun-ko:
Closing it. but we will consider these in v3

PR #221, opened 2024-12-27
slashing update etherfi viewer

solipsis:
* Removed methods that no longer have a non-deprecated equivalent
* Updated methods with that have a mostly equivalent call


seongyun-ko (approved):
lgtm

PR #222, opened 2024-12-30
[EigenLayer Slashing] (1) update EtherFiRestaker, (2) fixes in EtherFiNode

seongyun-ko:
- **EtherFiNode**
  - at `splitBalanceInExecutionLayer`, convert the share amount to the underlying asset amount
  - at `_queueEigenpodFullWithdrawal`, convert the share amount to the underlying asset amount + withdraw the full amount if it is the last validator of the pod
- **EtherFiRestaker**
  - update according to EigenLayer upgrade
  - deprecate the contract variable `withdrawalRootToWithdrawal`. the upgrade will be performed when there is no queued withdrawal. 
  - keep `withdrawalRootsSet` to help track the queued withdrawals
  - use `quoteByFairValue` where needed
  - add `selClaimerFor`

### DIFF
--- a/src/EtherFiRestaker.sol
+++ b/src/EtherFiRestaker.sol
@@ -254,7 +254,7 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable, 
         if (info.elStrategy != IStrategy(address(0))) {
             uint256 restakedTokenAmount = getRestakedAmount(_token);
             restaked = liquifier.quoteByFairValue(_token, restakedTokenAmount); /// restaked & pending for withdrawals
-            unrestaking = getEthAmountInEigenLayerPendingForWithdrawals(_token);
+            unrestaking = liquifier.quoteByFairValue(_token, getEthAmountInEigenLayerPendingForWithdrawals(_token));
         }
         holding = liquifier.quoteByFairValue(_token, IERC20(_token).balanceOf(address(this))); /// eth value for erc20 holdings
         pendingForWithdrawals = getEthAmountPendingForRedemption(_token);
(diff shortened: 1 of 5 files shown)


src/EtherFiRestaker.sol
```diff
@@ -39,8 +40,8 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
     mapping(address => TokenInfo) public tokenInfos;
     
     EnumerableSet.Bytes32Set private withdrawalRootsSet;
-    mapping(bytes32 => IDelegationManager.Withdrawal) public withdrawalRootToWithdrawal;
 
+    IRewardsCoordinator public immutable rewardsCoordinator;
```
solipsis, line 44: I believe removing the mapping and then declaring this new variable as immutable causes a storage issue since the immutable variable won't be in storage. There will be a conflict for the next storage variable we add in a future upgrade

src/EtherFiRestaker.sol
```diff
@@ -373,22 +309,20 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
...
-        // strategies[0] = strategy;
-        // uint256[] memory sharesArr = new uint256[](1);
-        // sharesArr[0] = shares;
-
-        // params[0] = IDelegationManager.QueuedWithdrawalParams({
-        //     strategies: strategies,
-        //     shares: sharesArr,
-        //     withdrawer: address(this)
-        // });
+    function _queueWithdrawlsByShares(address _token, uint256 _shares) internal returns (bytes32[] memory) {
```
solipsis, line 312: typo Withdrawls -> Withdrawals

src/EtherFiRestaker.sol
seongyun-ko: makes sense. we can keep the mapping variable putting `DEPRECATED_` prefix

src/EtherFiRestaker.sol
seongyun-ko: fixed

src/EtherFiRestaker.sol
seongyun-ko: fixed

solipsis (approved):

PR #224, opened 2024-12-31
[EigenLayer Slashing] Mock

seongyun-ko:
- see `test_launch_validator` in `test/EtherFiNode.t.sol`

### DIFF
--- a/test/eigenlayer-contracts/EigenPod.sol
+++ b/test/eigenlayer-contracts/EigenPod.sol
@@ -1,0 +1,736 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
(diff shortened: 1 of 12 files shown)


test/eigenlayer-utils/Logger.t.sol
```diff
@@ -0,0 +1,295 @@
+// SPDX-License-Identifier: BUSL-1.1
```
solipsis, line 1: this seems unused?

test/eigenlayer-contracts/EigenPod.sol
solipsis, line 1: If we want to instantiate real eigenpod contracts, shouldn't we vendor their contracts properly so that this does not get out of sync with any updates they do?

seongyun-ko: agreed. can u make the changes?

PR #225, opened 2025-01-06
El slashing mock testing

solipsis:
* INCLUDES A SINGLE LINE CHANGE TO `EtherfiNode.sol` that improves mock testing setup
* Mocks for core eigenlayer contracts required for withdrawals
* Helper library for arrays to clean up our code frequently creating arrays of 1 item
* Tests for
  * withdrawing last unslashed validator
  * withdrawing last slashed validator
  * withdrawing last validator with rewards
  * withdrawing non-final validator

### DIFF
--- a/src/EtherFiNode.sol
+++ b/src/EtherFiNode.sol
@@ -581,8 +581,7 @@ contract EtherFiNode is IEtherFiNode, IERC1271 {
     function createEigenPod() public {
         if (eigenPod != address(0x0)) return; // already have pod
         IEigenPodManager eigenPodManager = IEigenPodManager(IEtherFiNodesManager(etherFiNodesManager).eigenPodManager());
-        eigenPodManager.createPod();
-        eigenPod = address(eigenPodManager.getPod(address(this)));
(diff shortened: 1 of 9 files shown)


shivam-ef:
I assume we need some more changes to test an intermediate slashed withdrawal and hence we didn't test that?

shivam-ef (approved):
LGTM

solipsis:
> I assume we need some more changes to test an intermediate slashed withdrawal and hence we didn't test that?

Its a little awkward because eigenpods now work in a per pod model, while a bunch of our legacy contracts work in a per validator model. So if we exit (on beacon chain) 2 validators, 1 slashed and 1 unslashed. And then we go to process the withdrawals on the execution layer, there is no good onchain way to attribute a specific number of shares to a specific validator. If we process the slashed validator first, it will withdraw 32 ether. Then when we exit the other validator it will withdraw the remaining balance (less than 32 ether). If we process in the other order, the opposite happens. Same end result. In a future protocol upgrade we should better match eigenlayers model

PR #234, opened 2025-02-19
rewards manager contract

vvalecha519:


src/RewardsManager.sol
```diff
@@ -0,0 +1,156 @@
...
+    /// @param _recipients Array of addresses to receive rewards
+    /// @param _amounts Array of reward amounts corresponding to each recipient
+    function processRewards(address _token, address[] calldata _recipients, uint256[] calldata _amounts) external {
+        if (_recipients.length != _amounts.length) {
+            revert("Array lengths must match");
+        }
+        if (!roleRegistry.hasRole(REWARDS_MANAGER_ADMIN, msg.sender)) {
+            revert("Caller must be admin");
+        }
+        if (lastProcessedBlock[_token] + CLAIM_DELAY > block.number) {
```
shivam-ef, line 74: claim delay is in seconds but other two are in blocks, claim delay should be 1 day in blocks = 7200

src/RewardsManager.sol
```diff
@@ -0,0 +1,156 @@
...
+            revert("Claim delay not met");
```
shivam-ef, line 75: Can use custom errors for gas savings but this is fine as well

src/RewardsManager.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    /// @notice Process and allocate rewards for multiple recipients
+    /// @dev Moves current pending rewards to claimable and sets new pending amounts.
+    ///      When removing a recipient, their entry in the next processRewards call
+    ///      should have amount = 0 to properly move their pending rewards to claimable.
+    /// 
+    /// @param _token The address of the reward token being distributed
+    /// @param _recipients Array of addresses to receive rewards
+    /// @param _amounts Array of reward amounts corresponding to each recipient
+    function processRewards(address _token, address[] calldata _recipients, uint256[] calldata _amounts) external {
```
shivam-ef, line 67: Are funds sent here before the process rewards function is called?

src/RewardsManager.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    function initialize() external initializer {
+        __Ownable_init();
+        __UUPSUpgradeable_init();
+    }
+    /// @notice Updates the recipient address for a specific earner's rewards
+    /// @dev Can be called by either the earner themselves or a rewards manager admin
+    /// @param _earner The address of the rewards earner
+    /// @param _recipient The new address to receive the earner's rewards
+    function updateRewardsRecipient(address _earner, address _recipient) external {
```
seongyun-ko, line 57: how will this contract scale to 100~1000s earners?

src/RewardsManager.sol
```diff
@@ -0,0 +1,175 @@
...
+    /// @notice Process and allocate rewards for multiple recipients
+    /// @dev Moves current pending rewards to claimable and sets new pending amounts.
+    ///      When removing a recipient, their entry in the next processRewards call
+    ///      should have amount = 0 to properly move their pending rewards to claimable.
+    /// 
+    /// @param _token The address of the reward token being distributed
+    /// @param _recipients Array of addresses to receive rewards
+    /// @param _amounts Array of reward amounts corresponding to each recipient
+    /// @param _blockNumber The block number of the last block rewards were distributed
+    function processRewards(address _token, address[] calldata _recipients, uint256[] calldata _amounts, uint256 _blockNumber) external {
```
seongyun-ko, line 77: should we name `_recipients` to `_earners`? not sure what you actually intended

src/RewardsManager.sol
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        lastProcessedAtBlock[_token] = block.number;
+        rewardsCalculatedToBlock[_token] = _blockNumber;
+        uint256 tokenBalance = IERC20(_token).balanceOf(address(this));
+        for (uint256 i = 0; i < _recipients.length; i++) {
+            totalClaimableRewards[_token][_recipients[i]] += totalPendingRewards[_token][_recipients[i]];
+            totalPendingRewards[_token][_recipients[i]] = _amounts[i];
+            totalRewardsToDistribute[_token] += _amounts[i];
+        }
```
seongyun-ko, line 98: i dont understand the math here. 

why do we have the `totalPendingRewrads` seperate from `totalClaimableRewards`

src/RewardsManager.sol
seongyun-ko: how about `distributeRewards` instead of `processRewards`

src/RewardsManager.sol
```diff
@@ -0,0 +1,175 @@
...
+            revert("No rewards to claim");
+        }
+        if(earnerToRecipient[msg.sender] != address(0)) {
+            recipient = earnerToRecipient[msg.sender];
+        }
+        totalClaimableRewards[_token][msg.sender] = 0;
+        totalRewardsToDistribute[_token] -= amountToClaim;
+
+        if(_token == ETH_ADDRESS) {
+            totalClaimableRewards[_token][msg.sender] = 0;
```
seongyun-ko, line 121: you already updated it above at line 117

src/RewardsManager.sol
seongyun-ko: nvm. i got it

src/RewardsManager.sol
```diff
@@ -0,0 +1,175 @@
...
+import {RoleRegistry} from "./RoleRegistry.sol";
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {OwnableUpgradeable} from "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import {UUPSUpgradeable} from "@openzeppelin-upgradeable/contracts/proxy/utils/UUPSUpgradeable.sol";
+
+/// @title RewardsManager
+/// @notice Manages the distribution and claiming of rewards for the EtherFi protocol
+/// @dev Implements reward distribution logic for both ETH and ERC20 tokens
+contract RewardsManager is IRewardsManager, OwnableUpgradeable, UUPSUpgradeable {
```
seongyun-ko, line 15: i think this will become a merkle contract with timelock for claim in future. wdyt? should we just go there rn?

PR #246, opened 2025-03-24
Eigenlayer Slashing Upgrade

solipsis:
This PR ports the eigenlayer slashing changes to the v2.49 version of the protocol

This PR also lays the groundwork for a new mock based testing strategy around the eigenlayer contracts due to the fragile nature of maintaining the forked eigenlayer tests

Audited by Certora

- schedule: https://app.safe.global/transactions/tx?safe=eth:0xcdd57D11476c22d265722F68390b036f3DA48c21&id=multisig_0xcdd57D11476c22d265722F68390b036f3DA48c21_0x2b988ec51621639753d002dbb94c77624d64cf2b31084e9e8d3a83a3a29f40f6
- execute: https://app.safe.global/transactions/tx?safe=eth:0xcdd57D11476c22d265722F68390b036f3DA48c21&id=multisig_0xcdd57D11476c22d265722F68390b036f3DA48c21_0x1ed61775a24d3268bebf47a48dff789b965d5eb8c0d4d875a56d7b424017256c


src/EtherFiNode.sol
```diff
@@ -615,70 +578,68 @@ contract EtherFiNode is IEtherFiNode, IERC1271 {
...
+        // 
+        // TODO: revisit for Pectra where a validator can have more than 32 ether
+        uint256 depositSharesToWithdraw;
 
-        IDelegationManager delegationManager = IEtherFiNodesManager(etherFiNodesManager).delegationManager();
+        if (numAssociatedValidators() == 1) {
+            require(IEigenPod(eigenPod).activeValidatorCount() == 0, "ACTIVE_VALIDATOR_EXISTS");
+            depositSharesToWithdraw = depositShares[0];
+        } else {
+            uint256 eigenLayerBeaconStrategyShare = delegationManager.beaconChainETHStrategy().underlyingToShares(32 ether);
```
seongyun-ko, line 618: @etherfi-protocol/auditor 

Here is a finding by @solipsis 
```
What: call to a contract that doesn't exist delegationManager.beaconChainETHStrategy()
This function returns an IStrategy representing native beacon eth. Despite this returning the interface type, this is the 1 exception in eigenlayer's system and there is actually no contract deployed at this address. So even though it is declared as that interface, it doesn't actually implement the interface. This causes a failure when trying to call one of the interfaces methods underlayingToShares()

Fix: I believe the fix here is trivial. BeaconETH deposit shares (this is deposit shares not scaled shares) will always be 1 to 1 with wei so wei can simple hardcode 32 ether here instead of calling underlyingToShares()
```

seongyun-ko: resolved. @solipsis to confirm

solipsis: Yes this has been resolved and reviewed by the auditor

seongyun-ko (approved):
lgtm. we can merge after completing the upgrade

solipsis:
reverted `splitBalanceInExecutionLayer` update that will come in the next upgrade but was not deployed as part of this upgrade

seongyun-ko:
lgtm. the function `splitBalanceInExecutionLayer` is broken as of now, but not being used. it will be deprecated in next upgrade

PR #251, opened 2025-04-03
Feature: Add Venn Firewall to eETH

uri-chandler-ironblocks:
This PR introduces the Venn Firewall onto `eETH.sol`.

## Protected Methods
- `approve`
- `permit`

## Prerequisites
1. Update `foundry.toml` to use Solidity `0.8.25`
2. Replace
  2.1. `IERC20Upgradeable.sol` with `IERC20UpgradeablePayable.sol`
  2.2 `draft-IERC20PermitUpgradeable.sol` with `draft-IERC20PermitUpgradeablePayable.sol`
  2.3 ...both of which are identical versions, the only difference being that `approve` and `permit` are now `payable`
3. Mark `eETH.approve()` and `eETH.permit()` as `payable`

## Venn Integration
1. Install [firewall-consumer-v2](https://github.com/ironblocks/firewall-consumer-v2) as a dependency
2. Make the `eETH.sol` contract inherit from `ProxyVennFirewallConsumer.sol`
3. Add the `firewallProtected` modifier to protected methods
4. Add a new `initializeFirewallAdmin` method to allow the EtherFi team to set a new Firewall Admin after deploying the new eETH implementation

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "lib/solady"]
 	path = lib/solady
 	url = https://github.com/Vectorized/solady
+[submodule "lib/ironblocks/firewall-consumer-v2"]
+	path = lib/ironblocks/firewall-consumer-v2
+	url = https://github.com/ironblocks/firewall-consumer-v2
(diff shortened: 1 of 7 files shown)


seongyun-ko:
why

PR #261, opened 2025-05-02
Deterministic Contract Deployment

seongyun-ko:
This script provides a standardized, easy-to-review deployment flow using a deterministic contract deployment flow with Create2. It logs all deployment parameters and addresses in a structured JSON file for easy review and reproducibility. `EtherFiRedemptionManager` is used as an example.

## Security Guarantee
By using CREATE2 with a fixed salt and bytecode, the deployed contract address is fully predictable and verifiable before upgrade, ensuring no surprises or hidden logic.

## Reviewability
All constructor arguments and deployment metadata are automatically saved and human-readable, making audits and reviews straightforward and transparent for any contract deployed with this script.


shivam-ef (dismissed):

shivam-ef:
LGTM LFG!!!!

solipsis:
Looks good. One thing to maybe consider. Would we want the salt to be `ether.fi-{commitSlug}`? i.e. `ether.fi-abc123`

Could be nice to know exactly which commit each deployment should come from

seongyun-ko:
> Looks good. One thing to maybe consider. Would we want the salt to be `ether.fi-{commitSlug}`? i.e. `ether.fi-abc123`
> 
> Could be nice to know exactly which commit each deployment should come from

Good idea. I updated the script to use the commit hash to generate the salt with right padding

PR #262, opened 2025-05-07
Add recovery function in eETH and weETH for ETH, ERC20 and ERC721 tokens

shivam-ef:


solipsis:
I don't see anything wrong but I also don't like blanket rug ability this adds the contracts for an asset as widely trusted as ours. For a rescue operation like this, I think it should be a single function, hardcoded to recover exactly the target amount of the specific asset we are saving and only be allowed to run exactly once. I think it is easier to explain the the community if we get called out. If someone calls out the currently proposed upgrade I think we could lose a lot of credibility as being trustworthy.

I also think we need to prepare the full script that will upgrade, perform the rescue, and then remove the added function

shivam-ef:
> I don't see anything wrong but I also don't like blanket rug ability this adds the contracts for an asset as widely trusted as ours. For a rescue operation like this, I think it should be a single function, hardcoded to recover exactly the target amount of the specific asset we are saving and only be allowed to run exactly once. I think it is easier to explain the the community if we get called out. If someone calls out the currently proposed upgrade I think we could lose a lot of credibility as being trustworthy.
> 
> I also think we need to prepare the full script that will upgrade, perform the rescue, and then remove the added function

I agree. What do you think @seongyun-ko?

seongyun-ko:
> I don't see anything wrong but I also don't like blanket rug ability this adds the contracts for an asset as widely trusted as ours. For a rescue operation like this, I think it should be a single function, hardcoded to recover exactly the target amount of the specific asset we are saving and only be allowed to run exactly once. I think it is easier to explain the the community if we get called out. If someone calls out the currently proposed upgrade I think we could lose a lot of credibility as being trustworthy.
> 
> I also think we need to prepare the full script that will upgrade, perform the rescue, and then remove the added function

sure, i agree that we should not add any rug ability to our contracts.

but, what rug ability does it have?
these contracts will be able to recover only for assets that are not supposed to be held by the contract.

PR #277, opened 2025-07-30
Add verification script that uses implementation addresses directly

Gnome101:
The previous script checked the proxies and fetched the implementation address from the proxy. This one uses the implementation address directly, and checks the immutable variables.


pankajjagtapp:
V3 Prelude is done. Need to close this PR

PR #278, opened 2025-08-13
feat: add EL-triggerable batch exit support (EIP-7002) in EtherFiNodesManager

shyam-patel-kira:
## Summary

This PR implements Execution Layer-triggerable exits (EIP-7002) in the EtherFi smart contract by adding a new `batchWithdrawalRequests` method in `EtherFiNodesManager.sol`. This method allows authorized operators to submit one or more withdrawal requests to an EigenPod in a single transaction, enabling both full exits and partial withdrawals via the EIP-7002 predeploy mechanism.

### Notes

- Added a new Role `ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE` to control which operators can trigger EL exits.
- Update eigenPod interface (hopefully non-breaking!)

### DIFF
--- a/src/EtherFiNodesManager.sol
+++ b/src/EtherFiNodesManager.sol
@@ -8,6 +8,7 @@ import "@openzeppelin-upgradeable/contracts/security/ReentrancyGuardUpgradeable.
 import "@openzeppelin-upgradeable/contracts/security/PausableUpgradeable.sol";
 import "@openzeppelin/contracts/proxy/beacon/BeaconProxy.sol";
 import "./interfaces/IAuctionManager.sol";
+import "./eigenlayer-interfaces/IEigenPod.sol";
 import "./interfaces/IEtherFiNode.sol";
 import "./interfaces/IEtherFiNodesManager.sol";
 import "./interfaces/IProtocolRevenueManager.sol";
@@ -42,6 +43,7 @@ contract EtherFiNodesManager is
     bytes32 public constant ETHERFI_NODES_MANAGER_ADMIN_ROLE = keccak256("ETHERFI_NODES_MANAGER_ADMIN_ROLE");
     bytes32 public constant ETHERFI_NODES_MANAGER_EIGENLAYER_ADMIN_ROLE = keccak256("ETHERFI_NODES_MANAGER_EIGENLAYER_ADMIN_ROLE");
     bytes32 public constant ETHERFI_NODES_MANAGER_CALL_FORWARDER_ROLE = keccak256("ETHERFI_NODES_MANAGER_CALL_FORWARDER_ROLE");
+    bytes32 public constant ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE = keccak256("ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE");
 
     //-------------------------------------------------------------------------
     //-----------------------------  Admin  -----------------------------------
@@ -93,6 +95,57 @@ contract EtherFiNodesManager is
 
     function startCheckpoint(uint256 id) external onlyEigenlayerAdmin whenNotPaused {
         IEtherFiNode(etherfiNodeAddress(id)).startCheckpoint();
+    }
+
+    /**
+     * @notice Forwards EIP-7002 withdrawal requests to the EigenPod, supporting single or batch requests.
+     * @dev Access: only addresses with ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE.
+     * @dev Pausable + nonReentrant for safety.
+     * @param pod The EigenPod address that owns the validators.
+     * @param requests An array of WithdrawalRequest, where:
+     *        - requests[i].pubkey is the 48-byte BLS pubkey
+     *        - requests[i].amountGwei == 0 means "full exit"; >0 means partial to pod
+     * @custom:fee You MUST send sufficient ETH in msg.value to cover the EIP-7002 predeploy fee
+     *             for ALL requests in this batch (fee updates per block). Overpay is okay.
+     */
+    function batchWithdrawalRequests(
+        address pod,
+        IEigenPod.WithdrawalRequest[] calldata requests
+    ) external payable whenNotPaused nonReentrant onlyELTriggerRole()
+    {
+        // ---------- checks ----------
+        if (pod == address(0)) revert("pod=0");
+        uint256 n = requests.length;
+        if (n == 0) revert("empty");
+
+        // Strict pubkey length check to avoid wasting fee on malformed requests.
+        for (uint256 i = 0; i < n; ) {
+            bytes memory pk = requests[i].pubkey;
+            // Compute the pubkey hash
+            bytes32 pkHash = keccak256(pk);
+
+            // Ensure the node exists for this pubkey hash
+            if (address(etherFiNodeFromPubkeyHash[pkHash]) == address(0)) {
+                revert("unknown pubkey");
+            }
+
+            unchecked { ++i; }
+        }   
+
+        // Ensure enough fee is sent *at the time of execution*.
+        // NOTE: The predeploy updates per block; callers should slightly overpay.
+        uint256 feePer = IEigenPod(pod).getWithdrawalRequestFee();
+        // unchecked mul; we already ensure n>0 and feePer is bounded by protocol economics.
+        uint256 required = feePer * n;
+        if (msg.value < required) revert("insufficient fee");
+
+        // ---------- interactions ----------
+        // Forward full msg.value to tolerate fee update between the view and the call.
+        // If predeploy accepts >= required and internally refunds/credits excess, that's fine.
+        IEigenPod(pod).requestWithdrawal{value: msg.value}(requests);
+
+        // in case if event needs to be added
+        // emit BatchWithdrawalRequestsForwarded(msg.sender, pod, n, msg.value);
     }
 
     function verifyCheckpointProofs(uint256 id, BeaconChainProofs.BalanceContainerProof calldata balanceContainerProof, BeaconChainProofs.BalanceProof[] calldata proofs) external onlyEigenlayerAdmin whenNotPaused {
@@ -259,6 +312,11 @@ contract EtherFiNodesManager is
 
     modifier onlyCallForwarder() {
         if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_CALL_FORWARDER_ROLE, msg.sender)) revert IncorrectRole();
+        _;
+    }
+    
+    modifier onlyELTriggerRole() {
+        if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE, msg.sender)) revert IncorrectRole();
         _;
     }
 
--- a/src/eigenlayer-interfaces/IEigenPod.sol
+++ b/src/eigenlayer-interfaces/IEigenPod.sol
@@ -97,21 +97,33 @@ interface IEigenPodTypes {
         int64 balanceDeltasGwei;
         uint64 prevBeaconBalanceGwei;
     }
+
+    /**
+     * @param pubkey the pubkey of the validator to withdraw from
+     * @param amountGwei the amount (in gwei) to withdraw from the beacon chain to the pod
+     * @dev Note that if amountGwei == 0, this is a "full exit request," and will fully exit
+     * the validator to the pod.
+     * For more notes on usage, see `requestWithdrawal`
+     */
+    struct WithdrawalRequest {
+        bytes pubkey;
+        uint64 amountGwei;
+    }
 }
 
 interface IEigenPodEvents is IEigenPodTypes {
     /// @notice Emitted when an ETH validator stakes via this eigenPod
-    event EigenPodStaked(bytes pubkey);
+    event EigenPodStaked(bytes32 pubkeyHash);
 
     /// @notice Emitted when a pod owner updates the proof submitter address
     event ProofSubmitterUpdated(address prevProofSubmitter, address newProofSubmitter);
 
     /// @notice Emitted when an ETH validator's withdrawal credentials are successfully verified to be pointed to this eigenPod
-    event ValidatorRestaked(uint40 validatorIndex);
+    event ValidatorRestaked(bytes32 pubkeyHash);
 
     /// @notice Emitted when an ETH validator's  balance is proven to be updated.  Here newValidatorBalanceGwei
     //  is the validator's balance that is credited on EigenLayer.
-    event ValidatorBalanceUpdated(uint40 validatorIndex, uint64 balanceTimestamp, uint64 newValidatorBalanceGwei);
+    event ValidatorBalanceUpdated(bytes32 pubkeyHash, uint64 balanceTimestamp, uint64 newValidatorBalanceGwei);
 
     /// @notice Emitted when restaked beacon chain ETH is withdrawn from the eigenPod.
     event RestakedBeaconChainETHWithdrawn(address indexed recipient, uint256 amount);
@@ -128,10 +140,17 @@ interface IEigenPodEvents is IEigenPodTypes {
     event CheckpointFinalized(uint64 indexed checkpointTimestamp, int256 totalShareDeltaWei);
 
     /// @notice Emitted when a validator is proven for a given checkpoint
-    event ValidatorCheckpointed(uint64 indexed checkpointTimestamp, uint40 indexed validatorIndex);
+    event ValidatorCheckpointed(uint64 indexed checkpointTimestamp, bytes32 indexed pubkeyHash);
 
     /// @notice Emitted when a validaor is proven to have 0 balance at a given checkpoint
     event ValidatorWithdrawn(uint64 indexed checkpointTimestamp, uint40 indexed validatorIndex);
+
+    /// @notice Emitted when a withdrawal request is initiated where request.amountGwei == 0
+    event ExitRequested(bytes32 indexed validatorPubkeyHash);
+
+    /// @notice Emitted when a partial withdrawal request is initiated
+    event WithdrawalRequested(bytes32 indexed validatorPubkeyHash, uint64 withdrawalAmountGwei);
+
 }
 
 /**
@@ -212,6 +231,50 @@ interface IEigenPod is IEigenPodErrors, IEigenPodEvents, ISemVerMixin {
         bytes[] calldata validatorFieldsProofs,
         bytes32[][] calldata validatorFields
     ) external;
+
+    /// @notice Allows the owner or proof submitter to initiate one or more requests to
+    /// withdraw funds from validators on the beacon chain.
+    /// @param requests An array of requests consisting of the source validator and an
+    /// amount to withdraw
+    /// @dev The withdrawal request predeploy requires a fee is sent with each request;
+    /// this is pulled from msg.value. After submitting all requests, any remaining fee is
+    /// refunded to the caller by calling its fallback function.
+    /// @dev This contract exposes `getWithdrawalRequestFee` to query the current fee for
+    /// a single request. If submitting multiple requests in a single block, the total fee
+    /// is equal to (fee * requests.length). This fee is updated at the end of each block.
+    ///
+    /// (See https://eips.ethereum.org/EIPS/eip-7002#fee-update-rule for details)
+    ///
+    /// @dev Note on beacon chain behavior:
+    /// - Withdrawal requests have two types: full exit requests, and partial exit requests.
+    ///   Partial exit requests will be skipped if the validator has 0x01 withdrawal credentials.
+    ///   If you want your validators to have access to partial exits, use `requestConsolidation`
+    ///   to change their withdrawal credentials to compounding (0x02).
+    /// - If request.amount == 0, this is a FULL exit request. A full exit request initiates a
+    ///   standard validator exit.
+    /// - Other amounts are treated as PARTIAL exit requests. A partial exit request will NOT result
+    ///   in a validator with less than 32 ETH balance. Any requested amount above this is ignored.
+    /// - The actual amount withdrawn for a partial exit is given by the formula:
+    ///   min(request.amount, state.balances[vIdx] - 32 ETH - pending_balance_to_withdraw)
+    ///   (where `pending_balance_to_withdraw` is the sum of any outstanding partial exit requests)
+    ///   (Note that this means you may request more than is actually withdrawn!)
+    ///
+    /// @dev Note that withdrawal requests CAN FAIL for a variety of reasons. Failures occur when the request
+    /// is processed on the beacon chain, and are invisible to the pod. The pod and predeploy cannot guarantee
+    /// a request will succeed; it's up to the pod owner to determine this for themselves. If your request fails,
+    /// you can retry by initiating another request via this method.
+    ///
+    /// Some requirements that are NOT checked by the pod:
+    /// - request.pubkey MUST be a valid validator pubkey
+    /// - request.pubkey MUST belong to a validator whose withdrawal credentials are this pod
+    /// - If request.amount is for a partial exit, the validator MUST have 0x02 withdrawal credentials
+    /// - If request.amount is for a full exit, the validator MUST NOT have any pending partial exits
+    /// - The validator MUST be active and MUST NOT have initiated exit
+    ///
+    /// For further reference: https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-process_withdrawal_request
+    function requestWithdrawal(
+        WithdrawalRequest[] calldata requests
+    ) external payable;
 
     /**
      * @dev Prove that one of this pod's active validators was slashed on the beacon chain. A successful
@@ -294,12 +357,12 @@ interface IEigenPod is IEigenPodErrors, IEigenPodEvents, ISemVerMixin {
         bytes calldata validatorPubkey
     ) external view returns (ValidatorInfo memory);
 
-    /// @notice This returns the status of a given validator
+    /// @notice Returns the validator status for a given validator pubkey hash
     function validatorStatus(
         bytes32 pubkeyHash
     ) external view returns (VALIDATOR_STATUS);
 
-    /// @notice This returns the status of a given validator pubkey
+    /// @notice Returns the validator status for a given validator pubkey
     function validatorStatus(
         bytes calldata validatorPubkey
     ) external view returns (VALIDATOR_STATUS);
@@ -315,6 +378,11 @@ interface IEigenPod is IEigenPodErrors, IEigenPodEvents, ISemVerMixin {
 
     /// @notice Returns the currently-active checkpoint
     function currentCheckpoint() external view returns (Checkpoint memory);
+
+    /// @notice Returns the current fee required to add a withdrawal request to the EIP-7002 predeploy.
+    /// @dev Note that the predeploy updates its fee every block according to https://eips.ethereum.org/EIPS/eip-7002#fee-update-rule
+    /// Consider overestimating the amount sent to ensure the fee does not update before your transaction.
+    function getWithdrawalRequestFee() external view returns (uint256);
 
     /// @notice For each checkpoint, the total balance attributed to exited validators, in gwei
     ///


src/EtherFiNodesManager.sol
```diff
@@ -95,6 +97,57 @@ contract EtherFiNodesManager is
...
+     * @custom:fee You MUST send sufficient ETH in msg.value to cover the EIP-7002 predeploy fee
+     *             for ALL requests in this batch (fee updates per block). Overpay is okay.
+     */
+    function batchWithdrawalRequests(
+        address pod,
+        IEigenPod.WithdrawalRequest[] calldata requests
+    ) external payable whenNotPaused nonReentrant onlyELTriggerRole()
+    {
+        // ---------- checks ----------
+        if (pod == address(0)) revert("pod=0");
```
vvalecha519: can you use the style of throwing and Error

src/EtherFiNodesManager.sol
```diff
@@ -95,6 +97,57 @@ contract EtherFiNodesManager is
...
+        uint256 n = requests.length;
+        if (n == 0) revert("empty");
+
+        // Strict pubkey length check to avoid wasting fee on malformed requests.
+        for (uint256 i = 0; i < n; ) {
+            bytes memory pk = requests[i].pubkey;
+            // Compute the pubkey hash
+            bytes32 pkHash = keccak256(pk);
```
vvalecha519: Can you use similar variable naming pubkeyHash

src/EtherFiNodesManager.sol
```diff
@@ -95,6 +97,57 @@ contract EtherFiNodesManager is
...
+     * @param requests An array of WithdrawalRequest, where:
+     *        - requests[i].pubkey is the 48-byte BLS pubkey
+     *        - requests[i].amountGwei == 0 means "full exit"; >0 means partial to pod
+     * @custom:fee You MUST send sufficient ETH in msg.value to cover the EIP-7002 predeploy fee
+     *             for ALL requests in this batch (fee updates per block). Overpay is okay.
+     */
+    function batchWithdrawalRequests(
+        address pod,
+        IEigenPod.WithdrawalRequest[] calldata requests
+    ) external payable whenNotPaused nonReentrant onlyELTriggerRole()
```
vvalecha519: if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE, msg.sender)) revert IncorrectRole(); this is sytnax we have been following for checking role instead of using modifier

vvalecha519 (commented):
great! Some style changes + change you add test.

src/EtherFiNodesManager.sol
shyam-patel-kira: Oh okay, thanks!

src/EtherFiNodesManager.sol
shyam-patel-kira: Hey @vvalecha519, I did not notice pattern of throwing and Error() anywhere, Do we still want to change to it?

src/EtherFiNodesManager.sol
shyam-patel-kira: Updated this in my next commit

src/EtherFiNodesManager.sol
seongyun-ko: yes, 

e.g., from @vvalecha519 

```
if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE, msg.sender)) revert IncorrectRole();
```

seongyun-ko (commented):
q) what happens if the EoA holding the role `ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE` is compromised?

We should prevent the attacker from exiting all of our validators

e.g., Can we put the rate limit?
1. the EoA can operate normally within the configured rate limit
2. when we have to exit a lot more beyond the rate limit, our operating multi-sig shuold update the rate limit or fill the remaining cap.

can use this https://github.com/etherfi-protocol/smart-contracts/blob/master/src/BucketRateLimiter.sol

src/EtherFiNodesManager.sol
shyam-patel-kira: fixed this

shyam-patel-kira:
> Can we put the rate limit?
Added a `rateLimiter`, similar to one used in redemptionManager

vvalecha519 (commented):
we want to add rate limting for unrestaking as well

src/EtherFiNodesManager.sol
```diff
@@ -69,6 +73,10 @@ contract EtherFiNodesManager is
         _unpause();
     }
 
+    function __initRateLimiter() internal {
+        exitRequestsLimit = BucketLimiter.create(uint64(100), uint64(1));
```
solipsis: Since you never update this value, outside of this hardcoded initializer, please implement as an immutable that is set in the constructor

src/EtherFiNodesManager.sol
```diff
@@ -95,6 +103,76 @@ contract EtherFiNodesManager is
...
+     * @dev Pausable + nonReentrant for safety.
+     * @param pod The EigenPod address that owns the validators.
+     * @param requests An array of WithdrawalRequest, where:
+     *        - requests[i].pubkey is the 48-byte BLS pubkey
+     *        - requests[i].amountGwei == 0 means "full exit"; >0 means partial to pod
+     * @custom:fee You MUST send sufficient ETH in msg.value to cover the EIP-7002 predeploy fee
+     *             for ALL requests in this batch (fee updates per block). Overpay is okay.
+     */
+    function batchWithdrawalRequests(
+        address pod,
```
solipsis: I would drop this parameter entirely and no guarantee that it is a valid pod that we own. Instead look up the pod for each withdrawal request via the pubkey. This guarantees it is a pod we control and allows us to trigger exits from multiple pods at the same time

something like this
```
bytes32 pubkeyHash = calculateValidatorPubkeyHash(requests[i].pubkey)
address pod = etherFiNodeAddress(uint256(pubkeyHash)).getEigenPod();
```

src/EtherFiNodesManager.sol
```diff
@@ -95,6 +103,76 @@ contract EtherFiNodesManager is
...
+        if (!ok) revert ExitRateLimitExceeded();
+
+        // Strict pubkey length check to avoid wasting fee on malformed requests.
+        for (uint256 i = 0; i < n; ) {
+            bytes memory pubkey = requests[i].pubkey;
+            // Compute the pubkey hash
+            bytes32 pubkeyHash = keccak256(pubkey);
+
+            // Ensure the node exists for this pubkey hash
+            if (address(etherFiNodeFromPubkeyHash[pubkeyHash]) == address(0)) {
```
solipsis: this is wrong. Pubkey hash is not via keccak, it is via the SSZ serialization spec to match how eigenlayer computes it. you can use `calculateValidatorPubkeyHash`

src/EtherFiNodesManager.sol
```diff
@@ -95,6 +103,76 @@ contract EtherFiNodesManager is
...
+                revert UnknownNode();
+            }
+
+            unchecked { ++i; }
+        }   
+
+        // Ensure enough fee is sent *at the time of execution*.
+        // NOTE: The predeploy updates per block; callers should slightly overpay.
+        uint256 feePer = IEigenPod(pod).getWithdrawalRequestFee();
+        // unchecked mul; we already ensure n>0 and feePer is bounded by protocol economics.
```
solipsis: It is my understanding that all multiplications are now checked by default since 0.8.0 
So no need to worry about unchecked mul

src/interfaces/IEtherFiNodesManager.sol
```diff
@@ -105,6 +105,7 @@ interface IEtherFiNodesManager {
     event AllowedForwardedExternalCallsUpdated(bytes4 indexed selector, address indexed _target, bool _allowed);
     event AllowedForwardedEigenpodCallsUpdated(bytes4 indexed selector, bool _allowed);
     event FundsTransferred(address indexed nodeAddress, uint256 amount);
+    event BatchWithdrawalRequestsForwarded(address indexed operator, address indexed pod, uint256 requestCount, uint256 valueForwarded);
```
solipsis: Lets use a different name than operator for the first param. That is a very overloaded term in our protocol. This event might need to be reworked to support the other changes I mentioned above. I think its fine if we emit 1 event per withdrawalRequest instead of this batch

src/EtherFiNodesManager.sol
shyam-patel-kira: It's a struct, we're updating values in this struct often for e.g. cap, refillRate

src/EtherFiNodesManager.sol
shyam-patel-kira: okay thanks!

src/EtherFiNodesManager.sol
shyam-patel-kira: fixed this in my previous commit

src/EtherFiNodesManager.sol
solipsis: Ahh I didn't realize these functions were accepting storage pointers

src/EtherFiNodesManager.sol
```diff
@@ -95,6 +106,99 @@ contract EtherFiNodesManager is
...
+
+                unchecked { ++i; }
+            }
+        }
+
+        // Fee safety: require sufficient value at call time; forward full msg.value to tolerate fee bumps
+        // NOTE: The predeploy updates per block; callers should slightly overpay.
+        {
+            uint256 feePer = pod.getWithdrawalRequestFee();
+            if (msg.value < feePer * n) revert InsufficientWithdrawalFees();
```
solipsis: Just something to check, is this fee static for the entirety of a given TX, or is it possible it increases if say we exit 10 validators in this tx?. Don't want to fail if exiting validators causes the fee to rise slightly which would cause later exits to fail

test/prelude.t.sol
```diff
@@ -117,6 +130,7 @@ contract PreludeTest is Test, ArrayTestHelper {
         uint256 bidId;         // if none specified a new bid will be placed
         uint256 validatorSize; // if none specified default to 32 eth
         bool withdrawable;     // give the eigenpod "validatorSize" worth of withdrawable beacon shares
+        bytes pubkey;          // emulate validator from mainnet
```
solipsis: change comment -> if none specified a random pubkey is generated

test/prelude.t.sol
```diff
@@ -822,4 +838,393 @@ contract PreludeTest is Test, ArrayTestHelper {
...
+
+    // Resolve pod (and node) exactly as production does, using SSZ hash path.
+    function _resolvePod(bytes memory pubkey) internal view returns (IEtherFiNode node, IEigenPod pod) {
+        bytes32 pkHash = etherFiNodesManager.calculateValidatorPubkeyHash(pubkey);
+        IEtherFiNode etherFiNode = etherFiNodesManager.etherFiNodeFromPubkeyHash(pkHash);
+        pod = etherFiNode.getEigenPod();
+        require(address(pod) != address(0), "test: node has no pod");
+    }
+
+    function _sliceBytes(bytes[] memory arr, uint256 start, uint256 len) internal pure returns (bytes[] memory out) {
```
solipsis: solidity now has built in slicing operator. You can use that  instead of this helper

src/EtherFiNodesManager.sol
shyam-patel-kira: Within a block and txn  the fee for withdrawal remains the same, it updates every block in a way similar to EIP-1559. 
details: https://eips.ethereum.org/EIPS/eip-7002#fee-update-rule

test/prelude.t.sol
shyam-patel-kira: Can you point me to this? I found a library but it is better to just use this function instead of library since we'll need to install lib as dependency

test/prelude.t.sol
shyam-patel-kira: updated this

test/prelude.t.sol
shyam-patel-kira: turns out we cannot use the slice for memory array

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +133,100 @@ contract EtherFiNodesManager is
...
+        uint256 n = requests.length;
+        if (n == 0) revert EmptyWithdrawalsRequest();
+
+        {
+            uint64 units = SafeCast.toUint64(n);
+            bool ok = BucketLimiter.consume(exitRequestsLimit, units);
+            if (!ok) revert ExitRateLimitExceeded();
+        }
+
+        // ---- Resolve pod from first pubkey ----
```
seongyun-ko: can we combine this code block checking `requests[0]` 
and merge it into the below `for-loop`?

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +133,100 @@ contract EtherFiNodesManager is
...
+
+    function setExitRequestRefillPerSecond(uint256 refillPerSecond) external onlyAdmin() {
+        uint64 refill = SafeCast.toUint64(refillPerSecond);
+        BucketLimiter.setRefillRate(exitRequestsLimit, refill);
+    }
+
+    function canConsumeExitRequests(uint256 numRequests) external view returns (bool) {
+        return BucketLimiter.canConsume(exitRequestsLimit, SafeCast.toUint64(numRequests));
+    }
+
```
seongyun-ko, line 294: we need a GET function to return the fee for `requestWithdrawal`; 
return `getWithdrawalRequestFee` ?

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +133,100 @@ contract EtherFiNodesManager is
...
+    /**
+     * @notice Triggers EIP-7002 withdrawal requests, grouping by EigenPod automatically.
+     * @dev No `pod` param; we derive each pod from the validator pubkey using SSZ pubkey hash.
+     * @dev Access: only ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE, pausable, nonReentrant.
+     * @param requests Array of WithdrawalRequest:
+     *        - pubkey: 48-byte BLS pubkey
+     *        - amountGwei: 0 for full exit, >0 for partial to pod
+     * @custom:fee Send enough ETH to cover sum of (feePerPod * requestsForPod). Overpay is OK.
+     */
+    function requestWithdrawal(
```
seongyun-ko: seems this function accepts withdrawal requests only per pod/etherfi node?

we can release the constraint? so that it can process withdrawal requests across multiple pods as well

src/EtherFiNodesManager.sol
shyam-patel-kira: we want to emit the events after the call to `requestWithdrawal` is made

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +133,100 @@ contract EtherFiNodesManager is
...
+        IEigenPod.WithdrawalRequest[] calldata requests
+    ) external payable whenNotPaused nonReentrant
+    {
+        if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE, msg.sender)) revert IncorrectRole();
+        uint256 n = requests.length;
+        if (n == 0) revert EmptyWithdrawalsRequest();
+
+        {
```
seongyun-ko: let's add one internal function and call it from here for readability

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +133,100 @@ contract EtherFiNodesManager is
...
+            uint64 units = SafeCast.toUint64(n);
+            bool ok = BucketLimiter.consume(exitRequestsLimit, units);
```
seongyun-ko: ideally, the rate limit must be on the total amount of ETH being withdrawn

src/EtherFiNodesManager.sol
shyam-patel-kira: we made this trade-off because supporting multi-pod is much more gas intensive and requires more for loops, since we'd have to loop over each pod after differentiating them

src/EtherFiNodesManager.sol
shyam-patel-kira: it already exists in `IEigenPod` interface, do we want to add it to our contract as well?

src/EtherFiNodesManager.sol
solipsis: the timing of emitting events shouldn't matter

src/EtherFiNodesManager.sol
shyam-patel-kira: thanks for pointing out we already had a checker function that can be used here

src/EtherFiNodesManager.sol
shyam-patel-kira: merge this in my next commit

src/EtherFiNodesManager.sol
shyam-patel-kira: added a get function here

src/EtherFiNodesManager.sol
seongyun-ko: ok

src/EtherFiNodesManager.sol
```diff
@@ -69,6 +74,21 @@ contract EtherFiNodesManager is
...
+    }
+
+    function __initUnrestakingRateLimiter() external onlyAdmin() {
+        if (unrestakingLimit.lastRefill != 0) {
+            revert RateLimiterAlreadyInitialized();
+        }
+        // Use gwei as base unit like exit limiter: 172,800 ETH capacity, 2 ETH/sec refill
+        unrestakingLimit = BucketLimiter.create(uint64(172_800_000_000_000), uint64(2_000_000_000));
+    }
+
```
seongyun-ko: @solipsis plz can u check

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +142,136 @@ contract EtherFiNodesManager is
...
+        bytes32 pkHash0 = calculateValidatorPubkeyHash(requests[0].pubkey);
+        IEtherFiNode node0 = etherFiNodeFromPubkeyHash[pkHash0];
+        IEigenPod pod;        
+        {
+            if (address(node0) == address(0)) revert UnknownValidatorPubkey(); 
+            IEigenPod p0 = node0.getEigenPod();
+            if (address(p0) == address(0)) revert UnknownEigenPod();
+            pod = p0;
+        }
```
seongyun-ko: can u merge this block into the below block?

seongyun-ko: e.g.,

```
function requestWithdrawal(
    IEigenPod.WithdrawalRequest[] calldata requests
) external payable whenNotPaused nonReentrant {
    // ------------ role check ------------
    if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_EL_TRIGGER_EXIT_ROLE, msg.sender)) {
        revert IncorrectRole();
    }

    // ------------ basic checks ------------
    uint256 n = requests.length;
    if (n == 0) revert EmptyWithdrawalsRequest();

    uint256 totalExitGwei = getTotalEthRequested(requests);
    if (!canConsumeExitETH(totalExitGwei)) revert ExitRateLimitExceeded();

    IEtherFiNode node0;
    IEigenPod pod;
    uint256 feePer;

    // ------------ validate all requests in one loop ------------
    for (uint256 i = 0; i < n; ) {
        bytes32 pkHash = calculateValidatorPubkeyHash(requests[i].pubkey);
        IEtherFiNode node = etherFiNodeFromPubkeyHash[pkHash];
        if (address(node) == address(0)) revert UnknownValidatorPubkey();

        IEigenPod pi = node.getEigenPod();
        if (address(pi) == address(0)) revert UnknownEigenPod();

        if (i == 0) {
            // anchor pod from first request
            node0 = node;
            pod = pi;
            feePer = pod.getWithdrawalRequestFee();
        } else {
            if (pi != pod) revert PubkeysMapToDifferentPods();
        }

        emit ELWithdrawalRequestSent(
            msg.sender,
            address(pod),
            pkHash,
            requests[i].amountGwei,
            feePer
        );

        unchecked { ++i; }
    }

    // ------------ fee checks ------------
    if (msg.value < feePer * n) revert InsufficientWithdrawalFees();

    // ------------ external interaction ------------
    node0.requestWithdrawal{value: msg.value}(pod, requests);
}

```

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +142,136 @@ contract EtherFiNodesManager is
...
+
+    // Amount of ETH that refills per second once the bucket is not full
+    // for e.g. 2 ETH/second or 2_000_000_000 gwei/second rate would take 1 day to refill entirely 
+    function setExitETHRefillPerSecond(uint256 refillPerSecond) external onlyAdmin() {
+        uint64 refill = SafeCast.toUint64(refillPerSecond);
+        BucketLimiter.setRefillRate(exitRequestsLimit, refill);
+    }
+
+    // check for enough ETH left in remaining capacity
+    function canConsumeExitETH(uint256 totalEth) public view returns (bool) {
```
seongyun-ko: `totalEth` -> `amount`

shyam-patel-kira: fixed

src/EtherFiNodesManager.sol
shyam-patel-kira: I originally added the scope to avoid stack too deep, but it should fine now, merged the scoped code

vvalecha519 (commented):
This is great! Can we create role for unrestaking

seongyun-ko:
bugbot run

seongyun-ko:
bugbot run

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +134,118 @@ contract EtherFiNodesManager is
...
+
+        // ----------- external interaction --------
+        node.requestWithdrawal{value: msg.value}(requests);
+
+        // ------------ event emission -------------
+        for (uint256 i = 0; i < requests.length; ) {
+            bytes32 currentPubKeyHash = calculateValidatorPubkeyHash(requests[i].pubkey);
+            emit ValidatorWithdrawalRequestSent(address(pod), currentPubKeyHash, requests[i].pubkey);
+            unchecked { ++i; }
+        }
```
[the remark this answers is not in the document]
shyam-patel-kira: These are not risks

src/EtherFiNode.sol
```diff
@@ -140,7 +144,20 @@ contract EtherFiNode is IEtherFiNode {
...
     ///   For the general case of queuing a beaconETH withdrawal you can use queueETHWithdrawal instead.
-    function queueWithdrawals(IDelegationManager.QueuedWithdrawalParams[] calldata params) external onlyEigenlayerAdmin returns (bytes32[] memory withdrawalRoots) {
+    function queueWithdrawals(IDelegationManager.QueuedWithdrawalParams[] calldata params) external returns (bytes32[] memory withdrawalRoots) {
+        if (!roleRegistry.hasRole(ETHERFI_NODE_UNRESTAKER_ROLE, msg.sender)) revert IncorrectRole();
+        
+        // Calculate total beaconETH amount for rate limiting - only rate limit beaconETH strategy withdrawals
+        uint256 totalBeaconEth = 0;
+        for (uint256 i = 0; i < params.length; i++) {
+            for (uint256 j = 0; j < params[i].strategies.length; j++) {
+                if (params[i].strategies[j] == IStrategy(address(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0))) {
```
seongyun-ko: let's do as Copilot suggested. introduce `BEACON_ETH_STRATEGY_ADDRESS` as constant state

src/EtherFiNode.sol
```diff
@@ -140,7 +144,20 @@ contract EtherFiNode is IEtherFiNode {
...
+                    totalBeaconEth += params[i].depositShares[j];
+                }
+            }
+        }
+        etherFiNodesManager.consumeUnrestakingCapacity(totalBeaconEth);
```
seongyun-ko: I get it. but then we have to grant the role to all EtherFiNode. we have +20k of EtherFiNode instances.

hmmm. that is not impossible. is there any way to avoid it?

can we intro `queueWithdrawals` to EtherFiNodesManager contract and consume ratelimit there and forward the call to EtherFiNode?

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +143,129 @@ contract EtherFiNodesManager is
...
+        // ------------ fee checks ------------
+        if (msg.value < feePer * n) revert InsufficientWithdrawalFees();
+
+        // ------------ external interaction ----------
+        node0.requestWithdrawal{value: msg.value}(requests);
+    }
+
+    function getTotalEthRequested (IEigenPod.WithdrawalRequest[] calldata requests) internal pure returns (uint256) {
+        uint256 totalGwei; 
+        uint256 FULL_EXIT_GWEI = 2_048_000_000_000;
```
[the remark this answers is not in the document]
seongyun-ko: @shyam-patel-kira  let's declare it as a contract-level constant

src/EtherFiRateLimiter.sol
```diff
@@ -0,0 +1,204 @@
...
+
+import "@openzeppelin-upgradeable/contracts/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin-upgradeable/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin-upgradeable/contracts/security/PausableUpgradeable.sol";
+
+import "./interfaces/IEtherFiRateLimiter.sol";
+import "./interfaces/IRoleRegistry.sol";
+import "lib/BucketLimiter.sol";
+
+contract EtherFiRateLimiter is IEtherFiRateLimiter, Initializable, UUPSUpgradeable, PausableUpgradeable {
```
seongyun-ko, line 12: we need unit test on this new contract

src/EtherFiRateLimiter.sol
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    //--------------------------------------------------------------------------------------
+    //-----------------------------------  Migration  -----------------------------------
+    //--------------------------------------------------------------------------------------
+
+    /// @notice Migrate existing bucket limiter state during upgrade
+    /// @param id The limit identifier
+    /// @param legacyLimit The existing limit state to migrate
+    function migrateFromLegacyLimiter(bytes32 id, BucketLimiter.Limit calldata legacyLimit) external onlyAdmin {
```
seongyun-ko: where would we use this?

shyam-patel-kira: I believe during an upgrade but maybe @solipsis, can explain it better?

src/EtherFiNode.sol
shyam-patel-kira: okay, added a constant

shyam-patel-kira: Also changed where it was previously used as I used it here

src/EtherFiRateLimiter.sol
solipsis: Why did you add this function? this was not in the file I gave you?

solipsis: There is no migration required

src/EtherFiRateLimiter.sol
shyam-patel-kira: have added forks tests and added unit tests in my next commit

src/EtherFiRateLimiter.sol
shyam-patel-kira: weird, I directly used your code, anyway removed this as it is not required

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +134,118 @@ contract EtherFiNodesManager is
...
+        if (msg.value < pod.getWithdrawalRequestFee() * requests.length) revert InsufficientWithdrawalFees();
+
+        // ----------- external interaction --------
+        node.requestWithdrawal{value: msg.value}(requests);
+
+        // ------------ event emission -------------
+        for (uint256 i = 0; i < requests.length; ) {
+            bytes32 currentPubKeyHash = calculateValidatorPubkeyHash(requests[i].pubkey);
+            emit ValidatorWithdrawalRequestSent(address(pod), currentPubKeyHash, requests[i].pubkey);
+            unchecked { ++i; }
```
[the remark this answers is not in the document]
shyam-patel-kira: First issue is intentional and it is an Eigenlayer constraint. for Uint64 to overflow in gwei unit, one needs to trigger exits of 9M+ full 2048 ETH validators so this is not an issue

src/EtherFiNodesManager.sol
```diff
@@ -122,6 +134,118 @@ contract EtherFiNodesManager is
...
+                ? FULL_EXIT_GWEI
+                : uint256(requests[i].amountGwei);
+
+            totalGwei += gweiAmount;
+        }
+        return totalGwei;
+    }
+
+    function consumeUnrestakingCapacity(uint256 amount) external {
+        if (!roleRegistry.hasRole(ETHERFI_NODES_MANAGER_UNRESTAKER_ROLE, msg.sender)) revert IncorrectRole();
```
seongyun-ko: @solipsis this requires the caller (EFN) to has the permission

src/EtherFiNodesManager.sol
```diff
@@ -87,40 +101,197 @@ contract EtherFiNodesManager is
...
+        uint256 totalExitGwei = getTotalEthRequested(requests);
+        rateLimiter.consume(EXIT_REQUEST_LIMIT_ID, SafeCast.toUint64(totalExitGwei));
+
+        bytes32 pubKeyHash = calculateValidatorPubkeyHash(requests[0].pubkey);
+        IEtherFiNode node = etherFiNodeFromPubkeyHash[pubKeyHash];
+        IEigenPod pod = node.getEigenPod();
+
+        // submitting an execution layer withdrawal request requires paying a fee per request
+        if (msg.value < pod.getWithdrawalRequestFee() * requests.length) revert InsufficientWithdrawalFees();
+        node.requestExecutionLayerTriggeredWithdrawal{value: msg.value}(requests);
```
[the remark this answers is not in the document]
shyam-patel-kira: requests should belong to same pod is an eigenLayer constraint, if the txn reverts no capacity will be consumed since everything will be reverted

src/EtherFiRateLimiter.sol
```diff
@@ -0,0 +1,191 @@
...
+
+    //---------------------------------------------------------------------------
+    //----------------------------  ROLES  --------------------------------------
+    //---------------------------------------------------------------------------
+
+    bytes32 public constant ETHERFI_RATE_LIMITER_ADMIN_ROLE = keccak256("ETHERFI_RATE_LIMITER_ADMIN_ROLE");
+
+    //---------------------------------------------------------------------------
+    //-------------------------  Constants  ------------------------------------
+    //---------------------------------------------------------------------------
```
solipsis: drop this section

solipsis (dismissed):

src/EtherFiNode.sol
```diff
@@ -111,22 +105,25 @@ contract EtherFiNode is IEtherFiNode {
...
             uint32 slashableUntil = queuedWithdrawals[i].startBlock + EIGENLAYER_WITHDRAWAL_DELAY_BLOCKS;
             if (uint32(block.number) <= slashableUntil) continue;
             if (queuedWithdrawals[i].strategies.length != 1) continue;
-            if (queuedWithdrawals[i].strategies[0] != IStrategy(address(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0))) continue;
+            if (queuedWithdrawals[i].strategies[0] != IStrategy(BEACON_ETH_STRATEGY_ADDRESS)) continue;
 
             delegationManager.completeQueuedWithdrawal(queuedWithdrawals[i], tokens, receiveAsTokens);
+            anyWithdrawalsCompleted = true;
         }
+        if (!anyWithdrawalsCompleted) revert NoCompleteableWithdrawals(); // bad dev experience if function completes but nothing happened
```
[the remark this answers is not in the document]
seongyun-ko: no existing integration

test/prelude.t.sol
```diff
@@ -822,4 +884,790 @@ contract PreludeTest is Test, ArrayTestHelper {
...
+    function _requestsFromPubkeys(bytes[] memory pubkeys, uint64[] memory amountsGwei)
+        internal
+        pure
+        returns (IEigenPod.WithdrawalRequest[] memory reqs)
+    {
+        require(pubkeys.length == amountsGwei.length, "test: length mismatch");
+        reqs = new IEigenPod.WithdrawalRequest[](pubkeys.length);
+        for (uint256 i = 0; i < pubkeys.length; ++i) {
+            // NOTE: IEigenPod.WithdrawalRequest must match your interface type location
+            reqs[i] = IEigenPodTypes.WithdrawalRequest({pubkey: pubkeys[i], amountGwei: amountsGwei[i]});
```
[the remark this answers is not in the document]
shyam-patel-kira: this is not correct

src/EtherFiNodesManager.sol
```diff
@@ -87,40 +101,197 @@ contract EtherFiNodesManager is
...
+
+        for (uint256 i = 0; i < requests.length; ) {
+            bytes32 currentPubKeyHash = calculateValidatorPubkeyHash(requests[i].pubkey);
+            emit ValidatorWithdrawalRequestSent(address(pod), currentPubKeyHash, requests[i].pubkey);
+            unchecked { ++i; }
```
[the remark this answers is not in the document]
seongyun-ko: i agree with Copilot's suggestion. we dont really need to optimize for these. prefer better readability @shyam-patel-kira

script/DeployHoodiContracts.s.sol
```diff
@@ -0,0 +1,70 @@
...
+
+
+        console.log("Upgrading NodesManager proxy...");
+        IUpgradable(etherFiNodesManagerProxy).upgradeTo(address(newNodesManagerImpl));
+        console.log("  -> NodesManager proxy upgraded");
+
+        // 4) Upgrade EtherFiNode beacon via unified interface
+        console.log("Upgrading EtherFiNode beacon...");
+        IUpgradable(etherFiNodeProxy).upgradeTo(address(newNode));
+        console.log("  -> EtherFiNode beacon upgraded");
```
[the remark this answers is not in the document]
shyam-patel-kira: this is not relevant at this point

src/EtherFiNodesManager.sol
```diff
@@ -87,40 +101,196 @@ contract EtherFiNodesManager is
...
+        // submitting an execution layer withdrawal request requires paying a fee per request
+        if (msg.value < pod.getWithdrawalRequestFee() * requests.length) revert InsufficientWithdrawalFees();
+        node.requestExecutionLayerTriggeredWithdrawal{value: msg.value}(requests);
+
+        for (uint256 i = 0; i < requests.length; i++) {
+            bytes32 currentPubKeyHash = calculateValidatorPubkeyHash(requests[i].pubkey);
+            emit ValidatorWithdrawalRequestSent(address(pod), currentPubKeyHash, requests[i].pubkey);
+        }
+    }
+
```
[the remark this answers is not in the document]
shyam-patel-kira: this is fixed

src/EtherFiRateLimiter.sol
```diff
@@ -0,0 +1,184 @@
...
+    function isConsumerAllowed(bytes32 id, address consumer) external view returns (bool) {
+        return consumers[id][consumer];
+    }
+
+    /// @notice Checks if a rate limit exists
+    /// @param id The rate limit identifier
+    /// @return bool True if the rate limit has been initialized
+    function limitExists(bytes32 id) public view returns (bool) {
+        BucketLimiter.Limit memory limit = limits[id];
+        return limit.capacity != 0 || limit.remaining != 0 || limit.lastRefill != 0 || limit.refillRate != 0;
```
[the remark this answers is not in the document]
shyam-patel-kira: This works because, `last.refill` cannot be zero once the bucket is created

src/EtherFiNode.sol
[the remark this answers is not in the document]
shyam-patel-kira: not relevant

src/EtherFiNodesManager.sol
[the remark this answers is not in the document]
shyam-patel-kira: This is addressed

test/prelude.t.sol
```diff
@@ -89,24 +110,34 @@ contract PreludeTest is Test, ArrayTestHelper {
...
         vm.startPrank(roleRegistry.owner());
-        roleRegistry.grantRole(etherFiNodeImpl.ETHERFI_NODE_EIGENLAYER_ADMIN_ROLE(), address(etherFiNodesManager));
-        roleRegistry.grantRole(etherFiNodeImpl.ETHERFI_NODE_EIGENLAYER_ADMIN_ROLE(), address(stakingManager));
-        roleRegistry.grantRole(etherFiNodeImpl.ETHERFI_NODE_EIGENLAYER_ADMIN_ROLE(), eigenlayerAdmin);
-        roleRegistry.grantRole(etherFiNodeImpl.ETHERFI_NODE_CALL_FORWARDER_ROLE(), address(etherFiNodesManager));
         roleRegistry.grantRole(etherFiNodesManager.ETHERFI_NODES_MANAGER_ADMIN_ROLE(), admin);
         roleRegistry.grantRole(etherFiNodesManager.ETHERFI_NODES_MANAGER_CALL_FORWARDER_ROLE(), callForwarder);
         roleRegistry.grantRole(etherFiNodesManager.ETHERFI_NODES_MANAGER_EIGENLAYER_ADMIN_ROLE(), eigenlayerAdmin);
+        roleRegistry.grantRole(etherFiNodesManager.ETHERFI_NODES_MANAGER_EIGENLAYER_ADMIN_ROLE(), address(stakingManager));
+        roleRegistry.grantRole(etherFiNodesManager.ETHERFI_NODES_MANAGER_EIGENLAYER_ADMIN_ROLE(), address(stakingManager));
```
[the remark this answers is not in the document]
shyam-patel-kira: removed this

src/EtherFiNode.sol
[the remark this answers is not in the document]
shyam-patel-kira: repeated

src/EtherFiNode.sol
```diff
@@ -140,7 +137,7 @@ contract EtherFiNode is IEtherFiNode {
 
     /// @dev queue a withdrawal from eigenlayer. You must wait EIGENLAYER_WITHDRAWAL_DELAY_BLOCKS before claiming.
     ///   For the general case of queuing a beaconETH withdrawal you can use queueETHWithdrawal instead.
-    function queueWithdrawals(IDelegationManager.QueuedWithdrawalParams[] calldata params) external onlyEigenlayerAdmin returns (bytes32[] memory withdrawalRoots) {
+    function queueWithdrawals(IDelegationManager.QueuedWithdrawalParams[] calldata params) external onlyEtherFiNodesManager returns (bytes32[] memory withdrawalRoots) {
```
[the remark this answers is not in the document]
shyam-patel-kira: repeated wrong comment

src/EtherFiNode.sol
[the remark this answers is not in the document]
shyam-patel-kira: repeated comment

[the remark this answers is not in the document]
shyam-patel-kira: repeated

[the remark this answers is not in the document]
shyam-patel-kira: repeated

src/EtherFiNodesManager.sol
```diff
@@ -87,39 +102,194 @@
...
+            // Emit appropriate event based on whether this is a switch or consolidation
+            if (srcPkHash == targetPkHash) {
+                emit ValidatorSwitchToCompoundingRequested(address(pod), srcPkHash, requests[i].srcPubkey);
+            } else {
+                emit ValidatorConsolidationRequested(address(pod), srcPkHash, requests[i].srcPubkey, targetPkHash, requests[i].targetPubkey);
+            }
+            unchecked { ++i; }
+        }
+    }
```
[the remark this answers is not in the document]
shyam-patel-kira: repeated

PR #281, opened 2025-08-22
feat: add validators consolidation support EIP-7251

shyam-patel-kira:
## Summary

This PR implements EIP-7251 validator consolidation functionality for the etherFi protocol, following the same EIP-7002 withdrawal request patterns. This enables validator consolidation and switching from `0x01` to `0x02` (compounding withdrawal credentials)

### Notes:
- Combine multiple validators into single validators with up to 2,048 ETH
- Switch validators from `0x01` to `0x02` credentials for automatic reward compounding
- Up to 64x reduction in proof size and gas costs for large restakers

### DIFF
--- a/src/EtherFiNode.sol
+++ b/src/EtherFiNode.sol
@@ -157,7 +157,7 @@ contract EtherFiNode is IEtherFiNode {
             }
         }
         etherFiNodesManager.consumeUnrestakingCapacity(totalBeaconEth);
-        
+
         return delegationManager.queueWithdrawals(params);
     }
 
--- a/src/EtherFiNode.sol
+++ b/src/EtherFiNode.sol
@@ -146,15 +146,17 @@ contract EtherFiNode is IEtherFiNode {
     ///   For the general case of queuing a beaconETH withdrawal you can use queueETHWithdrawal instead.
     function queueWithdrawals(IDelegationManager.QueuedWithdrawalParams[] calldata params) external returns (bytes32[] memory withdrawalRoots) {
         if (!roleRegistry.hasRole(ETHERFI_NODE_UNRESTAKER_ROLE, msg.sender)) revert IncorrectRole();
-        // Calculate total amount for rate limiting (shares = wei, so each share * 1e18 = ether value)
-        uint256 totalAmount = 0;
+        
+        // Calculate total beaconETH amount for rate limiting - only rate limit beaconETH strategy withdrawals
+        uint256 totalBeaconEth = 0;
         for (uint256 i = 0; i < params.length; i++) {
-            for (uint256 j = 0; j < params[i].depositShares.length; j++) {
-                totalAmount += params[i].depositShares[j];
+            for (uint256 j = 0; j < params[i].strategies.length; j++) {
+                if (params[i].strategies[j] == IStrategy(address(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0))) {
+                    totalBeaconEth += params[i].depositShares[j];
+                }
             }
-        } 
-        // Check and consume unrestaking rate limit
-        etherFiNodesManager.consumeUnrestakingCapacity(totalAmount);
+        }
+        etherFiNodesManager.consumeUnrestakingCapacity(totalBeaconEth);
         
         return delegationManager.queueWithdrawals(params);
     }
--- a/src/EtherFiNode.sol
+++ b/src/EtherFiNode.sol
@@ -195,11 +195,11 @@ contract EtherFiNode is IEtherFiNode {
     //-------------------------------------------------------------------
     //-------------  Execution-Layer Triggered Consolidations  ----------
     //-------------------------------------------------------------------
-    function requestConsolidation(IEigenPod pod, IEigenPod.ConsolidationRequest[] calldata requests) external payable {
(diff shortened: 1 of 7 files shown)


src/EtherFiNode.sol
```diff
@@ -192,6 +192,16 @@ contract EtherFiNode is IEtherFiNode {
         pod.requestWithdrawal{value: msg.value}(requests);
     }
 
+    //-------------------------------------------------------------------
+    //-------------  Execution-Layer Triggered Consolidations  ----------
+    //-------------------------------------------------------------------
+    function requestConsolidation(IEigenPod pod, IEigenPod.ConsolidationRequest[] calldata requests) external payable {
```
solipsis: you should only ever get the pod via node.getEigenPod(). Not pass it in as a param

solipsis: I also think you should drop that param from the other PR as well

src/EtherFiNodesManager.sol
```diff
@@ -205,6 +205,72 @@ contract EtherFiNodesManager is
...
+     */
+    function requestConsolidation(
+        IEigenPod.ConsolidationRequest[] calldata requests
+    ) external payable whenNotPaused nonReentrant onlyAdmin
+    {
+        // ------------ checks ------------
+        uint256 n = requests.length;
+        if (n == 0) revert EmptyConsolidationRequest();
+
+        bytes32 pkHash0 = calculateValidatorPubkeyHash(requests[0].srcPubkey);
```
solipsis: I think this is way more complicated than it needs to be, I believe you can drop all of the code ensuring everything is tied to the same pod, eigenlayer's contracts already verify this

```
            // Ensure target has verified withdrawal credentials pointed at this pod
            bytes32 sourcePubkeyHash = _calcPubkeyHash(request.srcPubkey);
            bytes32 targetPubkeyHash = _calcPubkeyHash(request.targetPubkey);
            ValidatorInfo memory target = validatorPubkeyHashToInfo(targetPubkeyHash);
            require(target.status == VALIDATOR_STATUS.ACTIVE, ValidatorNotActiveInPod());
```

solipsis: Add again I believe you can drop the same checks from your first PR

src/interfaces/IEtherFiNodesManager.sol
```diff
@@ -110,6 +110,8 @@ interface IEtherFiNodesManager {
     event AllowedForwardedEigenpodCallsUpdated(bytes4 indexed selector, bool _allowed);
     event FundsTransferred(address indexed nodeAddress, uint256 amount);
     event ValidatorWithdrawalRequestSent(address indexed initiator, address indexed pod, bytes32 indexed validatorPubkeyHash, uint64 amountGwei, uint256 feePerRequest);
+    event ValidatorSwitchToCompoundingRequested(address indexed initiator, address indexed pod, bytes32 indexed validatorPubkeyHash, uint256 feePerRequest);
+    event ValidatorConsolidationRequested(address indexed initiator, address indexed pod, bytes32 indexed sourcePubkeyHash, bytes32 targetPubkeyHash, uint256 feePerRequest);
```
solipsis: swap initiator for targetPubkeyHash. It is way more likely to be a useful indexed value than the initiator

src/EtherFiNodesManager.sol
shyam-patel-kira: I believe they do not verify this but rather leave it to beacon chain to handle all the ownership related changes. EigenLayer trusts that if you submit an invalid request (validator doesn't belong to your pod), the beacon chain will simply **ignore/reject** it.

src/EtherFiNode.sol
```diff
@@ -146,15 +146,17 @@ contract EtherFiNode is IEtherFiNode {
...
+            for (uint256 j = 0; j < params[i].strategies.length; j++) {
+                if (params[i].strategies[j] == IStrategy(address(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0))) {
+                    totalBeaconEth += params[i].depositShares[j];
+                }
             }
-        } 
-        // Check and consume unrestaking rate limit
-        etherFiNodesManager.consumeUnrestakingCapacity(totalAmount);
+        }
+        etherFiNodesManager.consumeUnrestakingCapacity(totalBeaconEth);
```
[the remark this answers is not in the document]
shyam-patel-kira: this is intentional, eigenlayer has checks for these

src/EtherFiNodesManager.sol
```diff
@@ -184,6 +184,62 @@ contract EtherFiNodesManager is
...
+            }
+            unchecked { ++i; }   
+        }
+
+        // ------------ fee checks ------------
+        if (msg.value < feePer * n) revert InsufficientConsolidationFees();
+
+        // ------------ external interaction ----------
+        node0.requestConsolidation{value: msg.value}(requests);
+    }
```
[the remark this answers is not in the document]
shyam-patel-kira: This is intentional, eigenLayer has checks for these so we don't need to worry about validators being in the same pod

src/interfaces/IEtherFiNodesManager.sol
shyam-patel-kira: Updated these to more suitable params

PR #291, opened 2025-09-06
feat: new call forwarding strategy

shyam-patel-kira:
## Summary

Implemented a new granular call forwarding whitelist strategy for `EtherFiNodesManager` that adds user-specific permissions instead of the previous function-level permissions. This addresses security concerns where hot wallets with `CALL_FORWARDER_ROLE` had access to all whitelisted functions.

#### Storage Structure changes

```solidity 
// OLD - Function-level permissions
mapping(bytes4 => bool) public allowedForwardedEigenpodCalls;
mapping(bytes4 => mapping(address => bool)) public allowedForwardedExternalCalls;

 // NEW - User-specific permissions
mapping(address => mapping(bytes4 => bool)) public allowedForwardedEigenpodCalls;
mapping(address => mapping(bytes4 => mapping(address => bool))) public allowedForwardedExternalCalls;
```

#### Interface & Events Update
- Updated `IEtherFiNodesManager` interface to match new function signatures
- Consolidated events to user-specific variants only
- Updated view functions to accept user parameter

### why:
- Before: Any address with `CALL_FORWARDER_ROLE` could call ANY whitelisted function
- After: Each address can only call functions they're specifically granted permission for

### Remaining work:

- Add script for cleaning old storage use

### DIFF
--- a/src/EtherFiNodesManager.sol
+++ b/src/EtherFiNodesManager.sol
@@ -35,11 +35,9 @@ contract EtherFiNodesManager is
     //-----------------------------  Storage  -----------------------------------
     //---------------------------------------------------------------------------
     LegacyNodesManagerState private legacyState;
-    mapping(bytes4 => bool) public allowedForwardedEigenpodCalls; // Call Forwarding: functionSelector -> allowed
-    mapping(bytes4 => mapping(address => bool)) public allowedForwardedExternalCalls; // Call Forwarding: functionSelector -> targetAddress -> allowed
+    mapping(address => mapping(bytes4 => bool)) public allowedForwardedEigenpodCalls; // Call Forwarding: user -> functionSelector -> allowed
+    mapping(address => mapping(bytes4 => mapping(address => bool))) public allowedForwardedExternalCalls; // Call Forwarding: user -> functionSelector -> targetAddress -> allowed
     mapping(bytes32 => IEtherFiNode) public etherFiNodeFromPubkeyHash;
-    mapping(address => mapping(bytes4 => bool)) public userAllowedForwardedEigenpodCalls;
-    mapping(address => mapping(bytes4 => mapping(address => bool))) public userAllowedForwardedExternalCalls;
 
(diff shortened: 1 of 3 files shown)


src/EtherFiNodesManager.sol
```diff
@@ -414,12 +419,33 @@ contract EtherFiNodesManager is
...
             returnData[i] = IEtherFiNode(nodes[i]).forwardEigenPodCall(data[i]);
         }
     }
 
+    /// @notice Batch update the whitelist for external calls (convenience function)
+    /// @param user The address to grant/revoke permissions for
+    /// @param selectors Array of method selectors
+    /// @param targets Array of call targets for forwarded calls
+    /// @param allowed Array of enable/disable flags
+    function batchUpdateAllowedForwardedExternalCalls(
```
solipsis: no need for batch endpoint. This method will be called very rarely and we are already low on space

shyam-patel-kira: removed this

test/prelude.t.sol
```diff
@@ -392,6 +392,140 @@ contract PreludeTest is Test, ArrayTestHelper {
...
+        etherFiNodesManager.forwardExternalCall(toArray(etherFiNode), processClaimData, rewardsCoordinator);
+
+        // Test 5: Access control - only admin can update whitelists
+        vm.prank(user1);
+        vm.expectRevert(IEtherFiNode.IncorrectRole.selector);
+        etherFiNodesManager.updateAllowedForwardedExternalCalls(user2, transferSelector, usdc, true);
+
+        vm.prank(user1);
+        vm.expectRevert(IEtherFiNode.IncorrectRole.selector);
+        etherFiNodesManager.updateAllowedForwardedEigenpodCalls(user2, checkpointSelector, true);
```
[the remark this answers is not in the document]
shyam-patel-kira: good catch, fixed!

solipsis (approved):

PR #384, opened 2026-04-28
T1-6: Permissionless executeTasks()

0xpanicError:
> [!NOTE]
> **High Risk**
> Removes role gating from `EtherFiAdmin.executeTasks`, allowing any account to trigger state-changing reward/fee/validator/withdrawal processing once oracle consensus and sequencing/freshness checks pass. This increases operational exposure (griefing/DoS via public callers) and should be reviewed for reentrancy, gas, and trust assumptions around report validity.
> 
> **Overview**
> `EtherFiAdmin.executeTasks` is now **permissionless**: the executor role check was removed so any caller can process an oracle report once consensus is reached and the report passes the existing sequencing (`refSlotFrom`/`refBlockFrom`) and freshness wait checks.
> 
> Tests were updated to reflect this behavior by renaming the insufficient-role test to `test_executeTasks_permissionless` and asserting a role-less address can execute tasks and advance `lastHandledReportRefSlot`/`lastHandledReportRefBlock`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41a05fa6672951171b64b72c119c6b2c4f46746c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/EtherFiAdmin.sol
+++ b/src/EtherFiAdmin.sol
@@ -177,8 +177,6 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     }
 
     function executeTasks(IEtherFiOracle.OracleReport calldata _report) external {
-        if (!roleRegistry.hasRole(ETHERFI_ORACLE_EXECUTOR_TASK_MANAGER_ROLE, msg.sender)) revert IncorrectRole();
-
         bytes32 reportHash = etherFiOracle.generateReportHash(_report);
         uint32 current_slot = etherFiOracle.computeSlotAtTimestamp(block.timestamp);
         require(etherFiOracle.isConsensusReached(reportHash), "EtherFiAdmin: report didn't reach consensus");
--- a/test/EtherFiOracle.t.sol
+++ b/test/EtherFiOracle.t.sol
@@ -1347,7 +1347,7 @@ contract EtherFiOracleTest is TestSetup {
         etherFiOracleInstance.submitReport(wrongReport);
     }
 
-    function test_executeTasks_insufficientRole() public {
+    function test_executeTasks_permissionless() public {
         vm.prank(owner);
         etherFiOracleInstance.setQuorumSize(1);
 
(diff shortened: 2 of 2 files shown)


pankajjagtapp (approved):
lgtm

src/EtherFiAdmin.sol
seongyun-ko, line 1: it's significant implication and puts strong duty on Oracle

If Oracle committee is compromised:
1. what can happen in the worst-case
2. what we can do?

0xpanicError: We have added a bunch of sanity checks to ensure this. If in worst case oracle was comprised, the rebase cannot exceed to more than +-5% APR (which for a given report is 0.3 - 1.2 bps of TVL) so we can patch a fix in oracle and rectify accounting by next report (or over the course of a couple of days).

seongyun-ko: is that indeed all? let's brainstorm other paths under a compromised oracle. e.g.,

1. compromised oracle submits a report with `lastFinalizedWithdrawalRequestId = X` advancing well past current. is X bounded? can the attacker mark already-invalidated requests as valid (e.g. hacker-controlled NFT requests, dust requests we'd planned to invalidate) and then claim from those?
2. `_report.finalizedWithdrawalAmount = MAX_FINALIZED_WITHDRAWAL_AMOUNT_PER_DAY` every cycle for N days = drainage rate × N. what's our N before off-chain detection + UPGRADE_TIMELOCK rotation kicks in? days, weeks?
3. validator approval — if attacker submits attacker-controlled `validatorsToApprove`, the per-day cap limits volume but not identity. can attacker route validator funds to an attacker-controlled node operator?
4. protocol fees — capped at 20% of total rewards. if accruedRewards is forced to ±5% APR cap, fee is bounded. ok there.

i don't want the answer to be "lose ±5% TVL" if the actual answer is "drain `MAX_FINALIZED_WITHDRAWAL_AMOUNT_PER_DAY × N days` into attacker-controlled NFTs before we can rotate keys".

please walk through these in the PR description before merge.

0xpanicError: 1. you're right. the oracle submits `lastFinalizedWithdrawalRequestId` and `finalizedWithdrawalAmount`. We check liquidity exists against `finalizedWithdrawalAmount` but don't actually confirm that `lastFinalizedWithdrawalRequestId` would require `finalizedWithdrawalAmount` amount of locked liquidity. Should have on-chain check as well to counter against compromised oracle.

2. we check if liquidity even exists for this. combined with check to ensure `finalizedWithdrawalAmount` matches with finalized requests, compromised oracle can't submit arbitrary `finalizedWithdrawalAmount` values.

pankajjagtapp: 1. This check is added now.
2. Our N should be 3 days. But we have a settable variable with max value of 100k. Invalidation of requests is possible swiftly. We can mitigate such a situation before the funds are drained.
3. validatorsToApprove variable will be set to value only when we want to send ETH to vals. Otherwise will be kept zero to prevent malicious ETH movement. Morever, it is not possible to send ETH to random nodes. All ndoes are controlled by us.
4. okay

PR #389, opened 2026-04-30
yash/feat/min-max-withdraw-request

0xpanicError:
> [!NOTE]
> **Medium Risk**
> Changes withdrawal request validation in two core contracts, which can affect user-facing behavior and any integrations that submit very large withdrawals. Logic change is simple (constant bounds) but touches critical withdrawal entrypoints.
> 
> **Overview**
> Adds an upper bound of `1000 ether` for withdrawal requests across both the `PriorityWithdrawalQueue` and `WithdrawRequestNFT` flows, rejecting requests outside `[0.01 ether, 1000 ether]` (new `MAX_AMOUNT` / `MAX_WITHDRAW_AMOUNT`).
> 
> Updates and expands test coverage to assert min/max boundary behavior for eETH, weETH, and permit-based requests, and replaces the prior SD-6 sub-wei rounding test with an expectation that such tiny requests now revert (`InvalidWithdrawalAmount`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 961b91f70a40e49d4a71a98af428487dd370745b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/PriorityWithdrawalQueue.sol
+++ b/src/PriorityWithdrawalQueue.sol
@@ -34,6 +34,7 @@ contract PriorityWithdrawalQueue is
     //--------------------------------------------------------------------------------------
 
     uint96 public constant MIN_AMOUNT = 0.01 ether;
+    uint96 public constant MAX_AMOUNT = 1000 ether;
     uint256 private constant _BASIS_POINT_SCALE = 1e4;
 
     //--------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


pankajjagtapp (approved):

seongyun-ko (commented):
why do we do this? not against it. but asking why??

also isn't 1k too small? let's do 5k?

0xpanicError:
> why do we do this? not against it. but asking why??
> 
> also isn't 1k too small? let's do 5k?

this is for capital efficiency i think. if a user makes request for 5k ETH but as of now only 1k ETH is available. we withdraw 4k ETH and it stays idle until user claims request. with small caps, user has to make 5 1k requests which means 1k can be claimed as soon as it's finalized, and only 4k ETH stays idle until it gets added in liquidity.

seongyun-ko:
let's do 5k. 1k will be painful for whales.

also: codex flagged that membership-burn requests inherit the cap, which breaks existing flows. either carve out an exemption for the MembershipManager-originated path (or whichever caller does the burn) so it's not subject to the user-facing per-request cap, or update the logic in membership nft too.

pankajjagtapp:
> let's do 5k. 1k will be painful for whales.
> 
> also: codex flagged that membership-burn requests inherit the cap, which breaks existing flows. either carve out an exemption for the MembershipManager-originated path (or whichever caller does the burn) so it's not subject to the user-facing per-request cap, or update the logic in membership nft too.

Let's create another function which also splits the requests in multiple 1000 ETH NFTs @0xpanicError.

5k would not be ideal @seongyun-ko.
Will check the other flows on this one.

PR #416, opened 2026-05-20
fix/misc-changes

0xpanicError:
> [!NOTE]
> **Medium Risk**
> Medium risk due to widespread access-control/role-constant changes and an upgradeable storage change in `BNFT` that affects initialization and address wiring. Behavior is mostly equivalent (custom errors/renames), but misconfigured roles or missed initializer calls could break operational flows.
> 
> **Overview**
> Updates contracts and ops scripts to use the consolidated `RoleRegistry` tier roles (e.g., `ORACLE_OPERATIONS_ROLE`, `EXECUTOR_OPERATIONS_ROLE`, `HOUSEKEEPING_OPERATIONS_ROLE`, `EIGENPOD_OPERATIONS_ROLE`) instead of legacy `EOA_1..EOA_4`, including key gates in `EtherFiAdmin`, `EtherFiNodesManager`, `EtherFiRestaker`, and `CumulativeMerkleRewardsDistributor`.
> 
> Replaces many string-based `require` checks with *custom errors* across core contracts (`AuctionManager`, `BucketRateLimiter`, `DepositAdapter`, `EETH`, `EtherFiOracle`, `EtherFiRedemptionManager`, `EtherFiRewardsRouter`) and tightens a few callsite-specific revert reasons.
> 
> Makes `BNFT` fully upgradeable by moving constructor-set immutable addresses into storage and adding `initializeOnUpgrade` to set `etherFiNodesManagerAddress` post-upgrade, and adjusts `EETH` timed-pausing to require `onlySuperGuardian` instead of `onlyGuardian`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7ff1f922fa13871880d75e506126d4d0eb8e3c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pankajjagtapp (approved):

src/helpers/EtherFiViewer.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.13;
 
-import "@openzeppelin-upgradeable/contracts/proxy/utils/Initializable.sol";
```
pankajjagtapp, line 4: Dont think we are upgrading this contract

src/BNFT.sol
```diff
@@ -16,6 +16,10 @@ contract BNFT is ERC721Upgradeable, UUPSUpgradeable, OwnableUpgradeable {
     address public immutable stakingManagerAddress;
     address public immutable etherFiNodesManagerAddress;
 
+    error AddressZero();
```
pankajjagtapp: We arent upgrading these

0xpanicError: There is diff for BNFT in main branch. But it's only to deprecate old vars and make address immutable.

src/LiquidityPool.sol
```diff
@@ -4,11 +4,11 @@ pragma solidity ^0.8.13;
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts/utils/math/Math.sol";
-import "@openzeppelin-upgradeable/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgradeable/contracts/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
 
 import "./EtherFiRedemptionManager.sol";
```
pankajjagtapp: Remove this import

src/Liquifier.sol
```diff
@@ -81,7 +80,7 @@ contract Liquifier is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausab
 
     IPancackeV3SwapRouter pancakeRouter;
```
pankajjagtapp: Pancake code not removed yet

src/LiquidityPool.sol
0xpanicError: cant remove that, it's reference for deprecated var.

src/BNFT.sol
pankajjagtapp: Yeah, should skip

src/LiquidityPool.sol
pankajjagtapp: Can make it IEtherFiRedemptionManager though, right?

0xpanicError: technically yes but i think we should leave deprecated vars as it is even though underlying datatype is just address. because if we're okay changing data types we can replace all depreacted vars with empty uint[] arrys to block out slots.

pankajjagtapp: I agree but should remove this specifically because it had weirdly caused issues with compilation and verification on Etherscan earlier.

PR #444, opened 2026-05-29
docs(liquifier): stale stETH price-feed blocks deposit, not skipped (F7)

seongyun-ko:
## F7: clarify Liquifier stale-price-feed behavior (comment-only)

Documentation fix only. The comment above the stETH Chainlink price-feed check said "If price feed is stale, we skip the check" — which contradicts the code: a stale feed makes the deposit **revert** (`StalePriceFeed`), it is not skipped. Updated the comment to match the actual behavior (a stale feed blocks the stETH deposit).

No code/behavior change.

### Dropped from this PR: F5 (contract-scoped pause cooldown)
The `PausableUntil` contract-scoped-cooldown change (and its tests) has been removed from this PR — reverted to base. This PR is now scoped to the F7 comment only.


0xpanicError:
We want to keep the pausable until cooldown per address so that compromised guardian cannot spam pause repeatedly. Removing cooldown per address means a compromised EOA can start a cooldown for all guardian keys and then genuine hypernative key wouldn't be able to invoke pause until in case hack happens during cooldown.

PR #385, opened 2026-04-29
26Q2 - Security Upgrades

pankajjagtapp:
> [!NOTE]
> **Medium Risk**
> Touches governance/liquidity formal specs and many deployment/upgrade scripts aligned with role consolidation and adapter/oracle shape changes; on-chain impact depends on the separate contract upgrade batch, but wrong script constants or bytecode gates could block or mis-verify releases.
> 
> **Overview**
> This PR bundles **release and verification tooling** for the 26Q2 security-upgrades line, alongside broad **script/import refactors** that track the consolidated role model and new contract layout.
> 
> **Formal verification:** New Certora configs and CVL specs prove **LiquidityPool** peg/solvency wiring (**I7** `nonDecreasingRate` on guarded entry points, **I8** LP-buffer solvency after guarded writes, **I9** pooled-ether decomposition in a separate tautology spec) and **RoleRegistry** authority (**I6** owner/revokeAdmin/upgrade gates, no stray role mutation).
> 
> **CI & toolchain:** Forge tests on PRs targeting `pankaj/feat/security-upgrades*`; Foundry pinned to **v1.5.1** (avoids nightly EIP-3860 initcode failures on large harnesses); **`OP_RPC_URL`** added to the test env; **`foundry.lock`** gitignored.
> 
> **Build layout:** `remappings.txt` adds `@etherfi/`, `@tests/`, `@scripts/`, `@eigenlayer-libraries/`; scripts and libs switch from `../src/...` to those aliases. Eigenlayer library imports use the remapped package path.
> 
> **Deployment safety:** `ContractCodeChecker` gains **`assertByteCodeMatch`** — a reverting bytecode equality check that masks only address-derived immutables (and optional EIP-712 domain words), unlike the logging-only verifier.
> 
[...]

PR #445, opened 2026-05-29
P0 security hardening for #385: oracle price-move cap + restaker pause

seongyun-ko:
## P0 security hardening for #385

Two high-priority fixes from the security review. Each is small, isolated, and has a test. Base: `pankaj/feat/security-upgrades`.

---

### 1. Cap how much a single oracle report can move the weETH price (`EtherFiAdmin`)

**The problem, in plain terms:** the oracle periodically reports staking rewards, which moves the weETH↔eETH exchange rate. There's a safety check on this — but it's expressed as an *annual* rate (APR). The catch: the longer the gap a report covers, the bigger the one-time jump it's allowed to make while still looking fine on an annual basis.

**Example:** a normal daily report nudges the rate by ~0.01%. But a report that covers 60 days (say the oracle was paused and is catching up) is allowed to move the rate ~60× more in a single transaction — and still pass the APR check. A buggy or compromised oracle could exploit exactly this to inflate or crash the price in one shot. No other guard catches it, because a price change doesn't mint or burn any tokens.

**The fix:** add a hard ceiling — no single report may move total value by more than **2% of TVL**, no matter how long a window it covers. Normal reports are nowhere near this; anything bigger now has to be done deliberately by governance instead of sailing through automatically.

- 200 bps is a constant for now. If you'd rather make it adjustable, say so and I'll wire it as a governance setting with a fixed upper bound.
- A full report-level test belongs in the oracle test suite (it has the report builders). The arithmetic is simple; happy to add it if you point me at the harness.

---

### 2. Make the restaker's "pause" button actually do something (`EtherFiRestaker`)

**The problem:** `pauseContract()` flips a paused flag — but **not one function actually checks that flag.** So hitting pause did nothing: stETH transfers, ETH withdrawals, EigenLayer deposits/withdrawals all kept working. An operator who thought they'd stopped the contract during an incident hadn't.

**Example:** during an incident someone calls `pauseContract()` to freeze the restaker. An attacker (or a compromised ops key) calls `withdrawEther()` one block later — it still goes through, because nothing was checking "are we paused?".

**The fix:** every money-moving function now respects the pause. Also added `guardianPause()` so the Guardian (our Hypernative automation / emergency keys) can hit the brake instantly without gathering multisig signatures. Unpausing still requires the multisig (deliberate to turn off, easy to turn on — the intended asymmetry).

- Test `test_guardianPause_halts_fund_movement` confirms the pause now blocks transfers, that only the Guardian can fire it, and that resume is multisig-only. Passes on a mainnet fork.
---

### Test status
- New test passes on a mainnet fork.
- The pre-existing `EtherFiOracleTest::test_finalizeWithdrawalsWhenStale_*` failures (latest-block fork drift) reproduce identically on the untouched base branch — unrelated to this change.

---

> [!NOTE]
> **High Risk**
> Changes oracle report validation and the LiquidityPool share-rate rebase chokepoint, which directly affect eETH/weETH pricing; restaker pause wiring controls stETH/EigenLayer/ETH flows during incidents.
> 
> **Overview**
> This PR adds **defense-in-depth limits on how much TVL can move in a single rebase/report**, and makes **EtherFiRestaker pauses actually block fund flows**, aligned with the rest of the protocol.
> 
> **Rebase / oracle bounds:** `LiquidityPool.rebase` now rejects positive reward accrual above a **fixed 25 bps of pre-rebase TVL** (`RebaseExceedsPositiveCap`). `EtherFiAdmin` adds a separate **per-report negative (slashing) cap** (default **3 bps**, overridable via `setMaxNegativeRebaseBps` behind admin) so long-gap reports cannot pass the annualized APR check while still dropping TVL by a large absolute amount. Positive reward limits stay on the LP chokepoint; negative limits stay on report validation.
> 
> **Restaker pause:** Money-moving paths on `EtherFiRestaker` use `whenNotPaused`. The contract inherits **`PausableUntil`**: Guardian **`pauseContractUntil`**, multisig boolean pause/unpause, and `_requireNotPaused` also enforces the timed halt. `PausableUntil._pauseUntil` falls back to **`MIN_PAUSE_DURATION`** when duration was never set, avoiding a same-block no-op pause.
> 
> **Tests / harness:** Rebase-related tests use smaller accruals within the new cap; new tests cover the positive cap and restaker halts. `TestSetup` relaxes negative cap for generic suites via `setMaxNegativeRebaseBps(10_000)`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 019e84f82e0f08393aa95122e5b6ae88d0d1a93c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/core/LiquidityPool.sol
+++ b/src/core/LiquidityPool.sol
@@ -52,6 +52,10 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     uint256 public minWithdrawAmount;
     bool public escrowMigrationCompleted;
 
+    // Override for the per-rebase positive (reward) cap, in bps of TVL.
+    // 0 = use DEFAULT_MAX_POSITIVE_REBASE_BPS. Settable behind the operating timelock.
+    uint256 public maxPositiveRebaseBps;
+
     //--------------------------------------------------------------------------------------
     //-------------------------------------  IMMUTABLES  ----------------------------------
     //--------------------------------------------------------------------------------------
@@ -72,6 +76,13 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     //---------------------------------  CONSTANTS  ---------------------------------------
     //--------------------------------------------------------------------------------------
     uint256 public constant SHARE_UNIT = 1e18;
+
+    // Default cap on how far a single rebase may INCREASE TVL (rewards), in bps of TVL.
+    // 25 bps ≈ 1 month of reward accrual at 3% APR — generous for a normal report,
+    // tight enough that a buggy/compromised rebase caller cannot inflate the share rate
+    // arbitrarily. Overridable via `maxPositiveRebaseBps` behind the operating timelock.
+    uint256 public constant DEFAULT_MAX_POSITIVE_REBASE_BPS = 25;
+    uint256 private constant REBASE_BPS_DENOMINATOR = 10_000;
 
     //--------------------------------------------------------------------------------------
     //-------------------------------------  EVENTS  ---------------------------------------
@@ -94,6 +105,7 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     event ValidatorApproved(uint256 indexed validatorId);
     event ValidatorRegistrationCanceled(uint256 indexed validatorId);
     event Rebase(uint256 totalEthLocked, uint256 totalEEthShares);
+    event MaxPositiveRebaseBpsUpdated(uint256 bps);
     event ProtocolFeePaid(uint128 protocolFees);
     event WhitelistStatusUpdated(bool value);
     event MinWithdrawAmountSet(uint256 minWithdrawAmount);
@@ -113,6 +125,8 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     error InvalidValidatorSize();
     error InvalidAmountForShare();
     error InvalidRate();
+    error RebaseExceedsPositiveCap();
+    error InvalidMaxPositiveRebaseBps();
     error AlreadyMigrated();
     error MigrationNotComplete();
     error AlreadyRegistered();
@@ -438,11 +452,35 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     /// @notice Rebase by ether.fi
     function rebase(int128 _accruedRewards) public {
         if (msg.sender != address(membershipManager)) revert IncorrectCaller();
+
+        // Positive (reward) upper bound, enforced at the share-rate chokepoint regardless
+        // of who calls rebase. A single rebase cannot increase TVL by more than
+        // effectiveMaxPositiveRebaseBps() of pre-rebase TVL. Defense-in-depth alongside the
+        // oracle-side negative cap in EtherFiAdmin; the negative side is intentionally not
+        // re-checked here (the oracle path owns it and bounds it tighter).
+        if (_accruedRewards > 0) {
+            uint256 maxIncrease = (getTotalPooledEther() * effectiveMaxPositiveRebaseBps()) / REBASE_BPS_DENOMINATOR;
+            if (uint256(uint128(_accruedRewards)) > maxIncrease) revert RebaseExceedsPositiveCap();
+        }
+
         totalValueOutOfLp = uint128(int128(totalValueOutOfLp) + _accruedRewards);
 
         _checkMinAmountForShare();
 
         emit Rebase(getTotalPooledEther(), eETH.totalShares());
+    }
+
+    /// @notice Override the per-rebase positive (reward) cap, in bps of TVL.
+    /// @dev Operation-Timelock-gated. 0 resets to DEFAULT_MAX_POSITIVE_REBASE_BPS.
+    function setMaxPositiveRebaseBps(uint256 _bps) external onlyAdmin {
+        if (_bps > REBASE_BPS_DENOMINATOR) revert InvalidMaxPositiveRebaseBps();
+        maxPositiveRebaseBps = _bps;
+        emit MaxPositiveRebaseBpsUpdated(_bps);
+    }
+
+    function effectiveMaxPositiveRebaseBps() public view returns (uint256) {
+        uint256 v = maxPositiveRebaseBps;
+        return v == 0 ? DEFAULT_MAX_POSITIVE_REBASE_BPS : v;
     }
 
     /// @notice pay protocol fees including 5% to treaury, 5% to node operator and ethfund bnft holders
--- a/src/core/LiquidityPool.sol
+++ b/src/core/LiquidityPool.sol
@@ -52,10 +52,6 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     uint256 public minWithdrawAmount;
     bool public escrowMigrationCompleted;
 
-    // Override for the per-rebase positive (reward) cap, in bps of TVL.
-    // 0 = use DEFAULT_MAX_POSITIVE_REBASE_BPS. Settable behind the operating timelock.
-    uint256 public maxPositiveRebaseBps;
-
     //--------------------------------------------------------------------------------------
     //-------------------------------------  IMMUTABLES  ----------------------------------
     //--------------------------------------------------------------------------------------
@@ -77,11 +73,12 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     //--------------------------------------------------------------------------------------
     uint256 public constant SHARE_UNIT = 1e18;
 
-    // Default cap on how far a single rebase may INCREASE TVL (rewards), in bps of TVL.
-    // 25 bps ≈ 1 month of reward accrual at 3% APR — generous for a normal report,
-    // tight enough that a buggy/compromised rebase caller cannot inflate the share rate
-    // arbitrarily. Overridable via `maxPositiveRebaseBps` behind the operating timelock.
-    uint256 public constant DEFAULT_MAX_POSITIVE_REBASE_BPS = 25;
+    // Hard cap on how far a single rebase may INCREASE TVL (rewards), in bps of TVL.
+    // 25 bps ≈ 1 month of reward accrual at 3% APR — there is no legitimate reason for a
+    // single report to raise the rate by more than this, so it is a fixed invariant (not
+    // governance-configurable). Bounds a buggy/compromised rebase caller at the share-rate
+    // chokepoint regardless of the oracle-side checks.
+    uint256 public constant MAX_POSITIVE_REBASE_BPS = 25;
     uint256 private constant REBASE_BPS_DENOMINATOR = 10_000;
 
     //--------------------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


src/restaking/EtherFiRestaker.sol
```diff
@@ -397,12 +397,21 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
...
 
-    // Unpauses the contract
+    /// @notice Fast emergency halt callable by the Guardian (Hypernative / EOA keys).
+    /// @dev Pillar 2 (halt is cheap, redundant, parallel): the Guardian can stop fund
+    ///      movement immediately without assembling the multisig. Resume stays deliberate
+    ///      (multisig-only `unPauseContract`). This mirrors the token/LP halt model where
+    ///      the Guardian can pause but only the Operating Multisig can unpause.
+    function guardianPause() external onlyGuardian {
+        _pause();
+    }
```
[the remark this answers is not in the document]
seongyun-ko: Good catch — fixed in df243bd7. `undelegate()` now carries `whenNotPaused`: it queues withdrawal of all restaked assets, the same fund-flow category as `queueWithdrawals`, so the pause should halt it too. `delegateTo` is left ungated since it only changes delegation posture and queues/moves no funds. Test extended to assert `undelegate` reverts while paused.

src/oracle/EtherFiAdmin.sol
```diff
@@ -89,6 +89,18 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable, Rol
...
+    // long the report spans. `acceptableRebaseAprInBps` is annualized over elapsedTime,
+    // so a report covering a long window (e.g. a stale-report catch-up, or a maliciously
+    // back-dated `refSlotFrom`) permits a proportionally larger ABSOLUTE move while still
+    // passing the APR check. This hard ceiling bounds the per-report share-rate delta
+    // regardless of elapsedTime: a single report can never move TVL by more than this,
+    // so a compromised oracle quorum cannot inflate/deflate the exchange rate arbitrarily
+    // in one report. A legitimate move beyond this requires deliberate governance action.
+    // 200 bps = 2% of TVL; normal daily reports move ~1-3 bps, so this only trips on
+    // anomalous reports while still leaving generous headroom for multi-week catch-ups.
+    int256 public constant SINGLE_REPORT_REBASE_DELTA_CAP_BPS = 200;
```
0xpanicError: Let's set a min value per report to 3 bps which represents the max initial slashing penalty

0xpanicError: Rationale: initial slashing penalty is maxEffBalance / 4096, which is around 2.44 bps of tvl in case all validators inccur it. We should make this value configurable behing operatingTimelock so it can later be increased if we detect 18 days mid term slashing penalty event (can be detected way before 2 day timelock).

src/oracle/EtherFiAdmin.sol
```diff
@@ -410,6 +422,16 @@ contract EtherFiAdmin is Initializable, OwnableUpgradeable, UUPSUpgradeable, Rol
         }
         int256 absApr = (apr > 0) ? apr : - apr;
         if (absApr > acceptableRebaseAprInBps) return (false, "EtherFiAdmin: TVL changed too much");
+
+        // Absolute per-report cap, independent of elapsedTime. The APR check above is
+        // annualized, so a long-spanning report can pass it while still moving an
+        // outsized absolute amount of TVL in a single rebase. Bound the absolute move
+        // to SINGLE_REPORT_REBASE_DELTA_CAP_BPS of current TVL so no single report can
+        // shift the exchange rate beyond this, regardless of the window it covers.
+        int256 absRewards = (_report.accruedRewards > 0) ? int256(_report.accruedRewards) : -int256(_report.accruedRewards);
```
0xpanicError: Let's only compare accruedRewards per report with max_neg_rebase value (which we can set to 3 bps). to put invariant on upper-bound, better to put inside LP on functions expected to increase share rate and can set max upper-bound to something like 25 bps (that's 1 month of reward accrual at 3% APR).

src/restaking/EtherFiRestaker.sol
```diff
@@ -397,12 +397,21 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
...
         _pause();
     }
 
-    // Unpauses the contract
+    /// @notice Fast emergency halt callable by the Guardian (Hypernative / EOA keys).
+    /// @dev Pillar 2 (halt is cheap, redundant, parallel): the Guardian can stop fund
+    ///      movement immediately without assembling the multisig. Resume stays deliberate
+    ///      (multisig-only `unPauseContract`). This mirrors the token/LP halt model where
+    ///      the Guardian can pause but only the Operating Multisig can unpause.
+    function guardianPause() external onlyGuardian {
```
0xpanicError: why not use pausable until here?

0xpanicError: having the same function being called by 2 tiers of roles break the purpose of those roles.

src/oracle/EtherFiAdmin.sol
seongyun-ko: Done in dfac5251. Implemented exactly this on the negative side: a slashing cap on `accruedRewards` defaulting to **3 bps** (DEFAULT_MAX_NEGATIVE_REBASE_BPS), overridable via `setMaxNegativeRebaseBps` behind the operating timelock (0 = use default). So it ships tight at 3 bps and can be raised if a correlated/mid-term slashing event is detected. The old symmetric APR-only abs check is gone.

src/oracle/EtherFiAdmin.sol
seongyun-ko: Done in dfac5251. The negative side now only compares `accruedRewards` against the max-negative cap in EtherFiAdmin (default 3 bps). The positive/upper bound moved into `LiquidityPool.rebase` — the share-rate-increasing chokepoint — defaulting to **25 bps** (DEFAULT_MAX_POSITIVE_REBASE_BPS ≈ 1 month at 3% APR), settable behind the operating timelock via `setMaxPositiveRebaseBps`. Enforced there it holds regardless of which contract calls rebase. Test: `test_positiveRebaseCap_enforced`.

src/restaking/EtherFiRestaker.sol
seongyun-ko: Good call — switched to PausableUntil in dfac5251. The restaker now mirrors the protocol-wide model: `pauseContractUntil()` (onlyGuardian, auto-expiring 8h–30d with the per-pauser cooldown) + the boolean multisig `pauseContract()`, and a `whenNotHalted` modifier gates every fund-flow fn (incl. undelegate) on both. Replaced the OZ-only `guardianPause` I had. Tests updated (`test_guardianPauseUntil_halts_fund_movement`, `test_booleanPause_also_halts_fund_movement`).

src/core/LiquidityPool.sol
```diff
@@ -52,6 +52,10 @@ contract LiquidityPool is Initializable, OwnableUpgradeable, UUPSUpgradeable, Re
     uint256 public minWithdrawAmount;
     bool public escrowMigrationCompleted;
 
+    // Override for the per-rebase positive (reward) cap, in bps of TVL.
+    // 0 = use DEFAULT_MAX_POSITIVE_REBASE_BPS. Settable behind the operating timelock.
+    uint256 public maxPositiveRebaseBps;
```
0xpanicError: I don't think we need this value to be configurable. There's no way a single report should increase rate by more than 25 bps.

src/restaking/EtherFiRestaker.sol
```diff
@@ -143,7 +144,7 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
     /// @notice Transfer stETH to a recipient for instant withdrawal
     /// @param recipient The address to receive stETH
     /// @param amount The amount of stETH to transfer
-    function transferStETH(address recipient, uint256 amount) external {
+    function transferStETH(address recipient, uint256 amount) external whenNotHalted {
```
0xpanicError: can you use the `whenNotPaused` modifier used everywhere else for consistency? For contracts like this where PausableUpgradeable is inherited, override this internal function:

```
function _requireNotPaused() internal override view {
        _requireNotPausedUntil();
        super._requireNotPaused();
    }
```

src/restaking/EtherFiRestaker.sol
```diff
@@ -397,16 +398,41 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
...
 
-    // Pauses the contract
+    // Boolean pause (multisig). Indefinite until the multisig unpauses.
     function pauseContract() external onlyOperatingMultisig {
         _pause();
     }
 
-    // Unpauses the contract
+    // Unpauses the boolean pause (deliberate, multisig-only)
     function unPauseContract() external onlyOperatingMultisig {
```
0xpanicError: can you rename this: `unpauseContract()`

src/core/LiquidityPool.sol
seongyun-ko: Agreed — made it a hard constant in 019e84f8. `MAX_POSITIVE_REBASE_BPS = 25`, no setter/storage. (The negative/slashing cap in EtherFiAdmin stays timelock-configurable, since a real correlated-slashing event can legitimately exceed the initial-penalty default.) Rewrote the ~11 accounting tests that were doing 5%–100% rebases on tiny test TVL to use within-cap rebases.

src/restaking/EtherFiRestaker.sol
seongyun-ko: Done in 019e84f8 — exactly your pattern: overrode `_requireNotPaused()` to call `_requireNotPausedUntil()` + `super._requireNotPaused()`, dropped the bespoke `whenNotHalted`, and all fund-flow fns use the standard `whenNotPaused` again.

src/restaking/EtherFiRestaker.sol
seongyun-ko: Renamed to `unpauseContract()` in 019e84f8 (+ IEtherFiRestaker).

src/restaking/EtherFiRestaker.sol
seongyun-ko: Resolved by the PausableUntil switch: pausing is now two distinct functions on two tiers — `pauseContract()` (multisig, boolean) and `pauseContractUntil()` (Guardian, auto-expiring). No single function is callable by two role tiers anymore.

src/restaking/EtherFiRestaker.sol
```diff
@@ -397,16 +398,41 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
...
     }
 
+    /// @notice Fast, auto-expiring emergency halt callable by the Guardian (Hypernative /
+    ///         EOA keys). Mirrors the protocol-wide halt model (tokens, LP, redemption):
+    ///         the Guardian can stop fund movement immediately without assembling the
+    ///         multisig, the pause auto-expires (8h–30d) with a per-pauser cooldown so a
+    ///         wrong/compromised halt self-heals, and resume stays deliberate (multisig).
+    function pauseContractUntil() external onlyGuardian {
+        _pauseUntil();
+    }
```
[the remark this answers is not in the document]
seongyun-ko: Fixed in 019e84f8. `PausableUntil._pauseUntil` now falls back to `MIN_PAUSE_DURATION` (8h) when `pauseUntilDuration` is unset, so the emergency pause is always effective — no silent same-block expiry, and the cooldown is only burned on a real pause. (Chose the fallback over a revert-guard because much of the existing pause-until test surface relies on the duration-unset path.)

src/restaking/EtherFiRestaker.sol
```diff
@@ -397,16 +398,42 @@ contract EtherFiRestaker is Initializable, UUPSUpgradeable, OwnableUpgradeable,
...
+
+    /// @dev Fold the auto-expiring PausableUntil check into OZ's `_requireNotPaused`, so
+    ///      the standard `whenNotPaused` modifier (used on every fund-flow fn) halts on
+    ///      EITHER the boolean pause or the Guardian's auto-expiring pause — consistent
+    ///      with the rest of the protocol, no bespoke modifier needed.
+    function _requireNotPaused() internal view override {
+        _requireNotPausedUntil();
+        super._requireNotPaused();
+    }
```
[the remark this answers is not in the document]
0xpanicError: intended.

0xpanicError (approved):

PR #456, opened 2026-06-03
yash/fix/deprecate-oz-pausing

0xpanicError:
> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit c5936083fe594146f50580733613559197e46270. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/archive/EarlyAdopterPool.sol
+++ b/src/archive/EarlyAdopterPool.sol
@@ -5,9 +5,11 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/utils/math/Math.sol";
 import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
-import "@openzeppelin/contracts/security/Pausable.sol";
+// aliased to avoid a name clash with src/governance/utils/Pausable.sol when both are
+// pulled into the same compilation unit (e.g. TestSetup)
+import {Pausable as OZPausable} from "@openzeppelin/contracts/security/Pausable.sol";
 
-contract EarlyAdopterPool is Ownable, ReentrancyGuard, Pausable {
+contract EarlyAdopterPool is Ownable, ReentrancyGuard, OZPausable {
     using Math for uint256;
 
     struct UserDepositInfo {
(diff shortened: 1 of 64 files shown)


seongyun-ko (approved):
Approving. The contract changes and the storage layout are correct. Two things must be handled as part of the deployment, not the code merge.

## Critical — the pause functions are renamed, so anything calling the old names will fail

The pause entrypoints change:

- `pauseContract()` → `pause()`
- `unPauseContract()` → `unpause()`
- `pauseContractUntil()` → `pauseUntil()`
- `unpauseContractUntil()` → `unpauseUntil()`

(`setPauseUntilDuration(uint256)` is unchanged.)

The 4-byte selectors change along with the names, so any off-chain caller still pointing at the old names will hit a function that no longer exists and revert. This affects:

- Hypernative auto-pause actions (they send pre-built calldata to the old selectors)
- saved Safe transaction templates
- incident runbooks, including the no-UI chains where the call is pre-signed and can't be checked before broadcast

If this isn't updated, the emergency pause silently does nothing during an incident, while the dashboard may still show the call as sent.

Before the mainnet upgrade:

- Rebuild every Hypernative pause action and Safe template against the new function names, staged disabled, and cut them over at the upgrade.
- Simulate each auto-pause path against the new implementation on a fork.
- Update the pause runbooks and the entity-capability mapping.

Also note for integrators: the reentrancy revert is now `Reentrancy()` (was `ReentrancyGuardReentrantCall()`), and the pause events are now `Paused(address)` / `Unpaused(address)` (the event topic hash changed).

## High — confirm the pause caller actually holds the new role

Indefinite pause is now gated to the operating multisig, and timed pause to the guardian (super guardian for eETH/weETH). The role grants are not part of this PR. If the address that triggers the auto-pause doesn't hold the right role on each contract, the pause reverts on access control even after the function names are fixed — a second, independent failure on top of the rename.

Build a per-contract list of `{function, required role, does the caller hold it?}` and grant any missing roles in the same upgrade batch.

## Storage / rollout

Storage layout was verified byte-identical against the deployed contracts via `forge inspect`, so the upgrade itself is layout-safe. Before queuing, run the fork storage-integrity tests against the live proxies, and apply this PR before #457 (they are stacked and order matters).

PR #457, opened 2026-06-03
yash/fix/deprecate-oz-pausing-reentrancy

0xpanicError:
> [!NOTE]
> **High Risk**
> Touches reentrancy on deposits, withdrawals, and redemptions across live upgradeable proxies; layout must stay correct via DeprecatedOZReentrancyGuard, and transient guards depend on EIP-1153 support on the deployment chain.
> 
> **Overview**
> Migrates upgradeable core contracts from OpenZeppelin and custom **namespaced** reentrancy guards to Solady **`ReentrancyGuardTransient`**, which uses transient storage instead of persistent slots.
> 
> For proxies that previously inherited **`ReentrancyGuardUpgradeable`**, the PR adds **`DeprecatedOZReentrancyGuard`** (a 50-slot layout placeholder) in the same inheritance position, drops **`__ReentrancyGuard_init()`**, and layers **`ReentrancyGuardTransient`** where `nonReentrant` is still required (e.g. **Liquifier**, **AuctionManager**, **EtherFiNodesManager**, **EtherFiRedemptionManager**, **PriorityWithdrawalQueue**). **`LiquidityPool`** and **`WithdrawRequestNFT`** swap **`ReentrancyGuardNamespaced`** for the transient guard only. **`StakingManager`** keeps the deprecated OZ slot reservation but no longer pulls in an active guard in this diff.
> 
> Tests now expect **`ReentrancyGuardTransient.Reentrancy`**, remove namespaced **storage-slot** collision coverage, and adjust fork upgrade checks because the guard no longer has a plantable persistent slot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a10ddc2a44bc846bfbe4ef17b39cf59eb1a13f43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seongyun-ko (approved):
Approving. Stacked on #456 — apply after it.

The reentrancy guard moves to Solady's transient-storage version. Per-contract `nonReentrant` coverage is unchanged and cross-contract re-entry is still blocked, so there is no protection regression. Storage layout was verified byte-identical against the deployed contracts via the deprecated-slot placeholders.

Two notes for deployment:

- The transient guard uses transient storage only on mainnet and falls back to regular storage automatically on other chains — it does not break on chains that lack the transient-storage opcode. These contracts are mainnet-only today, so this is fine. Add a "mainnet / Cancun-only" note if any of them are ever deployed to another chain.
- Do not upgrade eETH/weETH while they are indefinitely paused. The paused flag changes storage location in this migration, so a pause set before the upgrade would not carry over and transfers would resume. Re-issue the pause after the upgrade if one is active.

Run the fork storage-integrity tests against the live proxies before queuing.

PR #478, opened 2026-06-20
fix: harden security-upgrade bytecode + storage verification

pankajjagtapp:
Fixes three verification gaps flagged in the security review of the 26Q2 security-upgrade scripts.

## Vuln 1 (HIGH) — bytecode verifier could never fail
`ContractCodeChecker` only `console.log`ed `Fail` on a mismatch and never reverted, so `verifyDeployedBytecode()` printed `[OK] ... matched` regardless of the actual code at the impl addresses. A wrong/stale/malicious impl address (only checked `!= address(0)`) would pass and be wired into the timelock batch.

- Add `assertByteCodeMatch(...)` — a hard-reverting gate. It requires equal length, then masks each contract's own 20-byte self-address (`UUPSUpgradeable.__self`) and tolerates a single caller-supplied 32-byte word (EETH's EIP-712 cached domain separator, read via `DOMAIN_SEPARATOR()` so it can't be a hiding spot). Any other differing byte reverts.
- Switch all 26 `verifyContractByteCodeMatch` calls in `transactions.s.sol` to `assertByteCodeMatch` (EETH passes its on-chain + local `DOMAIN_SEPARATOR()`).
- The legacy logging-only functions are kept for other callers.

Empirically, the only differences across all 26 impls were 464 self-address segments + one 32-byte EETH domain separator — exactly what the gate masks/allows.

## Vuln 2 (HIGH) — storage-layout fork coverage gap
The slot-scan integrity tests only covered 6 proxies. Added `UpgradeStorageIntegrityExtended.t.sol` extending the same pattern to the 8 previously-unverified upgraded proxies: **EETH, WeETH, EtherFiAdmin, Liquifier, StakingManager, EtherFiNodesManager, EtherFiRedemptionManager, MembershipManager**. Asserts 0 sequential storage drift across the impl swap, plus preserved `totalSupply()` (read through the new layout) for the value-bearing tokens. All 8 pass on a mainnet fork.

## Vuln 4 (MEDIUM) — verification not pinned to the release commit
`verifyDeployedBytecode` rebuilds each impl from the current working tree, so a match only proved "on-chain == my checkout". Added `_verifyReleaseCommit()` (requires `--ffi`) that pins HEAD to `GIT_COMMIT_SHA` and rejects a dirty tree. Gated by `VERIFY_GIT_COMMIT=true`; when skipped it logs a loud warning so the gap is never silent.

Also re-adds the `FORK_RPC_URL` override (documented in CLAUDE.md) so the hardened verification can run against the Tenderly testnet where the new impls are deployed.

## Testing
- `ContractCodeCheckerAssert.t.sol` (fork-free): 7/7 pass — gate accepts self-address-only and verified-word diffs; reverts on missing/wrong word, different contract, and empty on-chain code.
- `UpgradeStorageIntegrityExtended.t.sol` (mainnet fork): 8/8 pass, 0 drift.
- `forge build`: clean (0 errors).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes affect pre-broadcast upgrade verification and fork tests for many core proxies; incorrect masking in assertByteCodeMatch could false-pass or false-fail impl checks, but scope is verification tooling rather than on-chain upgrade logic.
> 
> **Overview**
> Closes security-review gaps in the **26Q2 security-upgrade** verification path so wrong impls, un-audited checkouts, and weak rate-limit checks cannot slip through silently.
> 
> **Bytecode verification (Vuln 1)** adds **`assertByteCodeMatch`** on `ContractCodeChecker` — a **reverting** equality check that masks only legitimate diffs (each impl’s **20-byte UUPS self-address** and, for EETH, a caller-supplied **EIP-712 domain separator** from `DOMAIN_SEPARATOR()`). The security-upgrade script replaces all **`verifyContractByteCodeMatch`** calls with this gate so bytecode mismatches **abort the script** instead of only logging “Fail”.
> 
> **Release pinning (Vuln 4)** adds **`_verifyReleaseCommit()`** (optional **`VERIFY_GIT_COMMIT=true`** + **`--ffi`**) to require **clean tree** and **HEAD == `GIT_COMMIT_SHA`**, with a loud warning when skipped. Fork selection prefers **`FORK_RPC_URL`** so verification can run where new impl bytecode exists.
> 
> **Storage layout (Vuln 2)** extends **`UpgradeStorageIntegrity.t.sol`** with sequential slot scans (and **`totalSupply`** checks where relevant) for **eight additional upgraded proxies**, inheriting **`SecurityUpgradesConstants`** instead of **`Deployed`**.
> 
> **Rate limits** tighten **`verifyOperatingConfig`** with capacity/refill checks against Constants, and post-upgrade flows add **`_assertExactBoundary`** for mint, burn, and stETH withdrawal buckets.
> 
> **`ContractCodeCheckerAssert.t.sol`** adds fork-free unit tests for the new gate, including shared-leading-byte address masking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e3536f300768341e3754923f40fdb9a2cc8c030. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pankajjagtapp:
Screenshot 2026-06-19 at 8 03 11 PM

0xpanicError (approved):

PR #483, opened 2026-07-14
ci: fix storage-layout check paths after src/ reorg

0xpanicError:
## Summary

The `Check storage layout` workflow has been failing on every run since the `src/` directory reorg: it still referenced the old flat paths (e.g. `src/EETH.sol`), so each step died with:

```
Error: Could not find source file for contract `EETH` at src/EETH.sol
```

This updates all 30 contract references in `.github/workflows/foundry-storage-check.yml` to their new locations (`src/core/`, `src/staking/`, `src/oracle/`, `src/archive/`, etc.). Contract names are unchanged — only file paths moved.

## Verification

- Every `path:ContractName` entry checked against the source tree (file exists and declares that contract)
- `forge inspect <entry> storage-layout` run locally for all 30 entries — all resolve and produce a layout

## Expected: the storage check is red on this PR itself

The [foundry-storage-check action](https://github.com/Rubilmax/foundry-storage-check) keys its baseline artifacts by contract path, so a PR run looks for `master.src_core_EETH.sol-EETH.json` — which can't exist until the new paths run on master once. This is the action's documented first-run behavior ("An error will appear at first run!" in its README) and there is no clean way around the bootstrap.

It self-heals on merge: on `push` events the action only generates and uploads baselines (it skips the comparison entirely — `if (context.eventName !== "pull_request") return`), so the merge push to master will go green and seed all 30 baseline artifacts. Every subsequent PR then compares normally. Note the check is already failing on master and is not a required check (#385 merged with it red).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/foundry-storage-check.yml
+++ b/.github/workflows/foundry-storage-check.yml
@@ -31,142 +31,142 @@ jobs:
       - name: Check storage layout for EETH
         uses: Rubilmax/foundry-storage-check@main
         with:
-          contract: src/EETH.sol:EETH
+          contract: src/core/EETH.sol:EETH
 
       - name: Check storage layout for WeETH
         uses: Rubilmax/foundry-storage-check@main
         with:
-          contract: src/WeETH.sol:WeETH
+          contract: src/core/WeETH.sol:WeETH
 
       - name: Check storage layout for LiquidityPool
         uses: Rubilmax/foundry-storage-check@main
         with:
-          contract: src/LiquidityPool.sol:LiquidityPool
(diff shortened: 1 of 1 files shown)


0xpanicError:
Note: the storage-layout check is expectedly red on this PR — baselines are keyed by contract path, and the new paths have no baseline until this runs on master once. The merge push uploads all 30 baselines (push runs skip comparison), so it goes green on merge and for all PRs after.

pankajjagtapp (approved):

